PR #1, opened 2021-03-24
Basic structure with user JWT auth

zsinx6:


jobbergate_api/apps/users/routers.py
```diff
@@ -0,0 +1,48 @@
+import typing
+
+from asyncpg.exceptions import UniqueViolationError
```
corydodt, line 3: I don't have any really strong opinions about this but i've been having a lot of success with the `databases` package, which is related to the `fastapi` family of apps. It provides a unified API that you can use with multiple database backends. only advantage i can actually point to is that you can do development with sqlite on your desktop before deploying to postgres on the backend. it is async (in fact i think it just includes asyncpg when you install it).

that might be valuable to you, or it might not

jobbergate_api/utils.py
```diff
@@ -0,0 +1,34 @@
...
+@click.command()
+def createsuperuser(username, email, password):
+    password = pwd_context.hash(password)
+    asyncio.run(create_super_user(username, email, password))
+    click.echo(f"User {username} created")
+
+
+cli.add_command(createsuperuser)
+
+if __name__ == "__main__":
```
corydodt, line 33: preferred style is just to create an entrypoint so you don't have to make this file executable, and it also makes discoverability of the command better.

then you don't need the last 2 lines at all

jobbergate_api/config.py
```diff
@@ -0,0 +1,14 @@
+from prettyconf import config
```
corydodt, line 1: we should just use [pydantic.BaseSettings](https://pydantic-docs.helpmanual.io/usage/settings/) which comes with a ton of validation and is highly configurable, and is already in your dependency stack.

jobbergate_api/main.py
```diff
@@ -0,0 +1,18 @@
+from fastapi import FastAPI
+from gino.ext.starlette import Gino
+
+from .config import settings
+from .routers import load_routers
+
+db = Gino()
+
+
+def get_app(db, database_url):
```
corydodt, line 10: you should separate app creation and db initialization for testability and to make it easier to refactor.

`fastapi` actually has event routers that are extremely useful for this kind of thing, see [init_database here](https://github.com/omnivector-solutions/license-manager/blob/main/src/licensemanager2/backend/main.py#L68)

corydodt: you might want to steal the `handler` function from the same file, because you'll need it in a lambda context, especially for api gateway

tests/apps/users/factories.py
```diff
@@ -0,0 +1,32 @@
+from jobbergate_api.apps.users.models import User
+
+
+class BaseFactory:
```
corydodt, line 4: Not sure I understand the purpose of this. this *looks* like a dataclass, and if so you could just use `pydantic.BaseModel` for that, but 
- you could also just use a unittest.mock.Mock object for that, but also
- why not just use the actual User object you already have in the app?

tox.ini
```diff
@@ -0,0 +1,25 @@
...
+skip_install = true
+setenv =
+    PYTHONPATH = {toxinidir}:{env:PYTHONPATH:}
+    TEST_ENV = true
+    TEST_DATABASE_URL=postgresql://postgres@localhost:5432/test_users_api
+    DATABASE_URL=postgresql://postgres@localhost:5432/users_api
+commands =
+    poetry install -v --no-root
+    poetry run pytest -vvv -s {posargs:tests} --cov jobbergate_api --cov-report=term-missing --cov-report=html
+    poetry run codecov
```
corydodt, line 17: why are there two coverage tools?

corydodt (changes requested):
- Please include docstrings in all modules, functions and methods (tests too)
- Please convert travis.yml to gha
- Please don't reuse `jobbergate_api` namespace. if this is a new project, it should have a new namespace. i suggest `jobbergateapi2`
- Not clear at this point whether this is pip-installable with only a pyproject.toml, could you do a little review of the docs on how to publish to pypi (keep in mind it will be a private pypi)

corydodt:
>     * Not clear at this point whether this is pip-installable with only a pyproject.toml, could you do a little review of the docs on how to publish to pypi (keep in mind it will be a private pypi)

forgot to mention an easy way to check:

1. create a virtualenv (preferably not the one you're already using, so you can make sure it's clean)
2. with the new env activated, `pip install -e .` or `pip install -e .[dev]` as appropriate
3. check that command-line tools you created from entrypoints or scripts do exist and can be run from the path
4. check `python -m jobbergateapi2` works, and/or run pytest

tox also installs its own environment, so tox tests passing might be enough to confirm the same thing. (make sure to run `tox -r` at some point to make sure the tox environment doesn't accidentally contain dependencies you forgot to add)

jobbergate_api/apps/users/schemas.py
```diff
@@ -0,0 +1,42 @@
+from datetime import datetime
```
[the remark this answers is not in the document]
zsinx6: It is currently using bcrypt
I searched about both, and bcrypt seems to be a little better: 
https://security.stackexchange.com/questions/4781/do-any-security-experts-recommend-bcrypt-for-password-storage/6415#6415
https://en.wikipedia.org/wiki/PBKDF2#Alternatives_to_PBKDF2

What do you think about it?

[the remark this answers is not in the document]
corydodt: My thoughts:

According to these links you posted,

- bcrypt has a small advantage in RAM usage,
- they have different characteristics when it comes to resisting specialized hardware attacks
- PBKDF2 is the standards-recommended solution,
- the stackexchange answer itself recommends PBKDF2

We're not terribly concerned about RAM usage in password workloads since we don't have that many actual (let alone concurrent) users. If we scale to the point where we can save money on RAM by switching algorithms, it looks like passlib itself lets us migrate pretty easily!

My recommendation is still PBKDF2 based on the above.

Additionally, consider:
- PBKDF2 parameters are really important. With some research, it looks to me like we should be doing something like:
    - PBKDF2-HMAC-SHA-512
    - 500,000 iterations (just as many as the system can take in a reasonable amount of time, this number goes up every year)
    - 64 byte (512 bit) derived key length with SHA512
- password length is really important, so we might want to prioritize long minimum passwords wherever we're allowing the user to set one
- Other hashing algorithms may or may not be superior to these two, but they need time to get proven.
- IF we're going to keep a password database, we should have a TOTP based 2fa solution (with U2F optional). This is currently the only reliable way to reduce unauthorized access to the system through a user account.
- BUT we should migrate away from doing our own auth with this system entirely. Prioritize building a system that integrates with an outside authorizer and accepts user metadata with the request, but no passwords.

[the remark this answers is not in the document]
corydodt: On that last point, it's obviously out of scope for this particular case, but it needs to come up the next time we're talking about a nontrivial upgrade to the jobbergate system.

Makefile
```diff
@@ -0,0 +1,33 @@
+test:
+	poetry run pytest -v
+
+lint:
+	poetry run pre-commit run -a -v
+
+install:
+	poetry install
+
+migrate:
```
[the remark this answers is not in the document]
zsinx6: The migrations helps to develop the application without interaction with the database, makes sense to put the migrations folder in the .gitignore for now, and keep it to develop only?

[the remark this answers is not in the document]
corydodt: I understand the purpose of database migrations, the question is, do we have to run one yet?

If we have no reason to run a migration yet, then building this prematurely will make it harder and more confusing for the next developer who has to figure out how to migrate the database for a new release.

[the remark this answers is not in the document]
corydodt: This might also be a candidate for like a doc/example/migration/ folder or something. Then you have it around to reference, but we're not obliged to test or maintain any migration code prematurely.

.github/workflows/test_on_push.yaml
```diff
@@ -0,0 +1,22 @@
...
+          python-version: '3.x'
+          architecture: 'x64'
+      - name: Python Poetry Action
+        uses: abatilo/actions-poetry@v2.1.0
+      - name: Install dependencies
+        run: |
+          pip install tox
+      - name: "tox lint"
+        run: |
+          tox -e lint
```
corydodt, line 22: Just run tox here, make sure all tests pass. If the other tests don't currently pass, we should address that asap after this PR

pyproject.toml
```diff
@@ -0,0 +1,55 @@
+[tool.poetry]
+authors = [
+    "zsinx6 <zsinx6@users.noreply.github.com>"
```
corydodt, line 3: Use your omnivector email here

jobbergateapi2/utils.py
```diff
@@ -0,0 +1,40 @@
+"""
```
corydodt, line 1: Can you just rename this whole file `createsuperuser.py` since it doesn't get used for anything else?

jobbergateapi2/utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+import click
+
+import jobbergateapi2.main as main
+from jobbergateapi2.apps.users.models import User
+from jobbergateapi2.apps.users.schemas import pwd_context
+from jobbergateapi2.config import settings
+
+
+@click.group()
```
corydodt, line 14: You don't have any other subcommands so you don't need these 3 lines at all.

jobbergateapi2/main.py
```diff
@@ -0,0 +1,34 @@
...
+from .config import settings
+from .routers import load_routers
+
+db = Gino()
+app = FastAPI()
+load_routers(app)
+
+
+@app.on_event("startup")
+async def init_database():
```
corydodt, line 17: good

jobbergateapi2/main.py
```diff
@@ -0,0 +1,34 @@
+"""
+Main file to startup the fastapi server
+"""
+from fastapi import FastAPI
+from gino.ext.starlette import Gino
+from mangum import Mangum
+
+from .config import settings
```
corydodt, line 8: Stylistically, please use absolute imports throughout, e.g. `from jobbergateapi2.config ..`

Relative imports can lead to surprises when you refactor.

jobbergateapi2/config.py
```diff
@@ -0,0 +1,20 @@
...
+class Settings(BaseSettings):
+    DATABASE_URL: str
+    TEST_DATABASE_URL: str
+    TEST_ENV: bool = Field(False)
+    SECRET_KEY: str
+    TOKEN_URL: str = Field("token")
+    ALGORITHM: str = Field("HS256")
+    ACCESS_TOKEN_EXPIRE_MINUTES: int = Field("30")
+
+    class Config:
```
corydodt, line 16: Ideally there should be an env prefix so it's clear what app these environment variables belong to.

jobbergateapi2/config.py
```diff
@@ -0,0 +1,20 @@
+"""
+Configuration file, sets all the necessary environment variables, it is better used with a .env file
+"""
+from pydantic import BaseSettings, Field
+
+
+class Settings(BaseSettings):
+    DATABASE_URL: str
+    TEST_DATABASE_URL: str
```
corydodt, line 9: Don't need to fix this right now but you can put regular expressions on these to help the operator validate.

corydodt (changes requested):

jobbergateapi2/config.py
zsinx6: Even when there will be only one app?

Makefile
```diff
@@ -0,0 +1,34 @@
+test:
+	poetry run pytest -v
```
jamesbeedy, line 2: how do you feel about adding a make command that checks for and installs poetry before you use it?

jamesbeedy: Or possibly just adding a getting started/readme that explains the process of getting up and running ....

default.env
```diff
@@ -0,0 +1,6 @@
+DATABASE_URL=postgresql://users_api:users_api@localhost/users_api
+TEST_DATABASE_URL=postgresql://users_api:users_api@localhost/test_users_api
```
jamesbeedy, line 2: I'm guessing these settings imply that you need to have a postgresql running locally when you develop or test the application. Can you add a bit to the readme about what the setup looks like here, 1) install postgresql, 2) create a db, etc etc.

default.env
```diff
@@ -0,0 +1,6 @@
...
+# openssl rand -hex 32
+SECRET_KEY=
```
jamesbeedy, line 4: Is this something that the user needs to set before testing, or is it ok to have an empty string default for `SECRET_KEY`?

zsinx6: it can be a random string, but that command generates a more secure/random string with larger size

Makefile
zsinx6: Just added a [README.md](https://github.com/omnivector-solutions/jobbergate-api-fastapi/blob/dev/README.md)

jobbergateapi2/apps/users/schemas.py
```diff
@@ -0,0 +1,58 @@
...
+        return f"{self.id}, {self.username}, {self.email}"
+
+
+class UserCreate(User):
+    """
+    Class used defines that a User have a password and garantee that it is hashed
+    """
+
+    password: str = Field(
+        None, min_length=8, max_length=32, description="A text value with length between 8 and 32 characters"
```
heitorPB, line 48: Could we raise the min and max length of passwords here?

zsinx6: The Field validates the field, so the exception is already raised here

heitorPB: I mean, can we have longer passwords?

zsinx6: How long?

heitorPB: AFAIK the recommended minimum password length is 12 chars, and the maximum is near 100. It is very easy to brute force all combinations of 10 ASCII chars.

jamesbeedy:
👍  LGTM. Let's get this bad boy merged. Send it.

PR #2, opened 2021-04-07
1186165980 Change database lib to databases

zsinx6:


jamesbeedy:
seeing 

```
CryptographyDeprecationWarning: int_from_bytes is deprecated, use int.from_bytes instead
    from cryptography.utils import int_from_bytes, int_to_bytes

-- Docs: https://docs.pytest.org/en/latest/warnings.html
``` 

in the `make test` output

Can we rid that warning pls?

jobbergateapi2/config.py
```diff
@@ -3,10 +3,11 @@
 """
 from pydantic import BaseSettings, Field
 
+_DB_RX = r"^(sqlite|postgres)://.+$"
```
jamesbeedy, line 6: 👍

jamesbeedy (approved):
Test pass on my local, runtime function seems to be 👍 . LGTM

jamesbeedy:
100% - thank you

jamesbeedy:
1186149976

jobbergateapi2/config.py
```diff
@@ -3,10 +3,11 @@
...
+
 
 class Settings(BaseSettings):
-    DATABASE_URL: str
-    TEST_DATABASE_URL: str
+    DATABASE_URL: str = Field("sqlite:///./sqlite.db?check_same_thread=true", regex=_DB_RX)
     TEST_ENV: bool = Field(False)
     SECRET_KEY: str
```
matheushent, line 12: Possibly usage here is [SecretStr](https://pydantic-docs.helpmanual.io/usage/types/#secret-types)

matheushent (approved):

PR #7, opened 2021-04-27
Job scripts - GET

zsinx6:
I have grouped some chunks of code in functions for the POST method, a great portion of it will likely be removed, so in this way it is easier to refactor it in the future.


jobbergateapi2/apps/job_scripts/routers.py
```diff
@@ -90,27 +96,15 @@ async def job_script_create(
...
         if member.name in default_template:
             contentfobj = s3_application_tar.extractfile(member)
             template_files["application.sh"] = contentfobj.read().decode("utf-8")
         if member.name in supporting_files:
-            match = [x for x in support_files_ouput if member.name in x]
+            match = [x for x in support_files_output if member.name in x]
             contentfobj = s3_application_tar.extractfile(member)
-            filename = support_files_ouput[match[0]][0]
+            filename = support_files_output[match[0]][0]
             template_files[filename] = contentfobj.read().decode("utf-8")
```
zsinx6, line 108: This is not ideal yet, and will likely be removed in the future, but the focus now is to keep it working like the original jobbergate-api first

jobbergateapi2/apps/job_scripts/routers.py
```diff
@@ -46,42 +46,48 @@ def inject_sbatch_params(job_script_data_as_string: str, sbatch_params: List[str
...
+            detail=f"Application with id={application_id} not found for user={current_user_id} in S3",
         )
-    application = Application.parse_obj(raw_application)
+    s3_application_tar = tarfile.open(fileobj=BytesIO(s3_application_obj["Body"].read()))
+    return s3_application_tar
 
+
+def render_template(template_files, param_dict_flat):
+    """
+    Use jinga2 to render the template as string.
```
jamesbeedy, line 71:
```suggestion
    Render the template as string using jinja2.
```

jamesbeedy (approved):
👍 from me. Nice work!

PR #11, opened 2021-05-20
Updates to work with frontend

zsinx6:
Changed the `User` fields to add `is_superuser`, `is_active` and changed `username` to `full_name`, and removed the `is_admin`.

Added CORS middleware

Created 2 new endpoints:
- `/users/{user_id}` to get a specific `User`
- `/user/me` to get the authenticated user information

### DIFF
--- a/jobbergateapi2/apps/users/routers.py
+++ b/jobbergateapi2/apps/users/routers.py
@@ -23,7 +23,8 @@ router = APIRouter()
 )
 async def users_list(p: Pagination = Depends()):
     """
-    Return all the users.
+    Return all the users with pagination.
+    The limit and skip are passed via query strings.
     """
     query = users_table.select().limit(p.limit).offset(p.skip)
     raw_users = await database.fetch_all(query)
@@ -39,7 +40,7 @@ async def users_list(p: Pagination = Depends()):
 )
 async def users_get(user_id: int = Query(...)):
     """
-    Return the user given it's id.
+    Return the user given its id.
     """
     query = users_table.select(users_table.c.id == user_id)
     raw_user = await database.fetch_one(query)
--- a/jobbergateapi2/apps/users/routers.py
+++ b/jobbergateapi2/apps/users/routers.py
@@ -1,11 +1,11 @@
 """
-Router for the User resource
+Router for the User resource.
 """
 import typing
 
-from fastapi import APIRouter, Depends, HTTPException
+from fastapi import APIRouter, Depends, HTTPException, Query, status
 
-from jobbergateapi2.apps.auth.authentication import validate_token
+from jobbergateapi2.apps.auth.authentication import get_current_user, validate_token
 from jobbergateapi2.apps.users.models import users_table
 from jobbergateapi2.apps.users.schemas import User, UserCreate
 from jobbergateapi2.compat import INTEGRITY_CHECK_EXCEPTIONS
@@ -17,19 +17,17 @@ router = APIRouter()
 
 @router.get(
     "/users/",
-    description="Endpoint to search users",
+    description="Endpoint to list users",
     response_model=typing.List[User],
     dependencies=[Depends(validate_token)],
 )
-async def users_search(p: Pagination = Depends()):
+async def users_list(p: Pagination = Depends()):
     """
-    Endpoint that requires authentication and is used to GET the users and returns using pagination
+    Return all the users.
     """
-    if p.q is None:
-        query = users_table.select().limit(p.limit).offset(p.offset)
-    else:
-        query = users_table.select().where(users_table.c.username == p.q).limit(p.limit).offset(p.offset)
-    users = await database.fetch_all(query)
+    query = users_table.select().limit(p.limit).offset(p.skip)
+    raw_users = await database.fetch_all(query)
+    users = [User.parse_obj(x) for x in raw_users]
     return users
 
 
@@ -42,10 +40,11 @@ async def users_create(user_data: UserCreate):
         try:
             query = users_table.insert()
             values = {
-                "username": user_data.username,
+                "full_name": user_data.full_name,
                 "email": user_data.email,
                 "password": user_data.hash_password(),
-                "is_admin": user_data.is_admin,
+                "is_superuser": user_data.is_superuser,
+                "is_active": user_data.is_active,
             }
             user_created_id = await database.execute(query=query, values=values)
 
--- a/jobbergateapi2/apps/users/routers.py
+++ b/jobbergateapi2/apps/users/routers.py
@@ -31,10 +31,43 @@ async def users_list(p: Pagination = Depends()):
     return users
 
 
+@router.get(
+    "/users/{user_id}",
+    description="Endpoint to get user given the id",
+    response_model=User,
+    dependencies=[Depends(validate_token)],
+)
+async def users_get(user_id: int = Query(...)):
+    """
+    Return the user given it's id.
+    """
+    query = users_table.select(users_table.c.id == user_id)
+    raw_user = await database.fetch_one(query)
+    if not raw_user:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"User {user_id=} not found",
+        )
+    user = User.parse_obj(raw_user)
+    return user
+
+
+@router.get(
+    "/user/me",
+    description="Endpoint to get current user",
+    response_model=User,
+)
+async def users_me(current_user: User = Depends(get_current_user)):
+    """
+    Return the authenticated user.
+    """
+    return current_user
+
+
 @router.post("/users/", description="Endpoint for user creation", dependencies=[Depends(validate_token)])
 async def users_create(user_data: UserCreate):
     """
-    Endpoint used to create new users using a user already authenticated
+    Create new users.
     """
     async with database.transaction():
         try:
--- a/jobbergateapi2/apps/users/routers.py
+++ b/jobbergateapi2/apps/users/routers.py
@@ -1,9 +1,9 @@
 """
 Router for the User resource.
 """
-import typing
+from typing import List, Optional
 
-from fastapi import APIRouter, Depends, HTTPException, Query, status
+from fastapi import APIRouter, Body, Depends, HTTPException, Query, status
 
 from jobbergateapi2.apps.auth.authentication import get_current_user, validate_token
 from jobbergateapi2.apps.users.models import users_table
@@ -18,7 +18,7 @@ router = APIRouter()
 @router.get(
     "/users/",
     description="Endpoint to list users",
-    response_model=typing.List[User],
+    response_model=List[User],
     dependencies=[Depends(validate_token)],
 )
 async def users_list(p: Pagination = Depends()):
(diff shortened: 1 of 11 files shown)


jobbergateapi2/tests/test_pagination.py
```diff
@@ -19,17 +19,15 @@ def test_pagination_with_default_values():
...
 
-    assert pagination.q is None
     assert pagination.limit == 10
-    assert pagination.offset == 0
+    assert pagination.skip == 0
 
 
 def test_pagination_with_custom_values():
     """
     Test the pagination with custom values
```
jamesbeedy, line 28:
```suggestion
    Test the pagination with custom values.
```

jobbergateapi2/tests/apps/users/test_routers.py
```diff
@@ -35,18 +35,17 @@ async def test_search(client):
 
 @pytest.mark.asyncio
 @database.transaction(force_rollback=True)
-async def test_search_with_pagination_limit_offset(client):
+async def test_list_with_pagination_limit_offset(client):
     """
-    Test if the pagination is working throught the endpoint with 2 users
+    Test if the pagination is working throught the endpoint with 2 users.
```
djkresta, line 40: "throught" --> "through"

jobbergateapi2/tests/apps/users/test_routers.py
```diff
@@ -55,11 +54,12 @@ async def test_search_with_pagination_limit_offset(client):
     assert count[0][0] == 2
 
 
-def test_search_without_results(client):
+@pytest.mark.asyncio
+async def test_list_without_results(client):
     """
-    Test when there is no user matching the search criteria
+    Test when there is no user.
```
djkresta, line 60: "Test listing functionality (using "/users/") when no users exist to list."

jobbergateapi2/tests/apps/users/test_routers.py
```diff
@@ -55,11 +54,12 @@ async def test_search_with_pagination_limit_offset(client):
...
     """
-    response = client.get("/users/?name=Some Name")
+    response = client.get("/users/")
 
     assert response.status_code == status.HTTP_200_OK
```
djkresta, line 64: Should we also check that nothing was returned?

jobbergateapi2/tests/apps/users/test_routers.py
```diff
@@ -71,22 +71,23 @@ def test_search_without_results(client):
 @database.transaction(force_rollback=True)
 async def test_create_user(client, user_data):
     """
-    Test user creation
+    Test user creation, should be active and not superuser.
```
djkresta, line 74: "Test the creation of a user who is active, but not a superuser."

jobbergateapi2/tests/apps/users/test_routers.py
```diff
@@ -99,15 +100,56 @@ async def test_create_user_duplication(client, user_data):
...
+    data = response.json()
+    assert data["email"] == user_data["email"]
+
+
+@pytest.mark.asyncio
+async def test_users_get_empty(client):
+    """
+    Try to get the user of id=1 which doesn't exists and will return 404.
+    """
+    response = client.get("/users/1")
```
djkresta, line 138: did we not just create a user with id=1 in the previous test?

jobbergateapi2/tests/apps/users/test_schemas.py
```diff
@@ -59,8 +59,8 @@ def test_create_user(user_data):
     user = User(**user_data)
 
     assert user.email == user_data["email"]
-    assert user.is_admin is False
-    assert user.username == user_data["username"]
+    assert user.is_superuser is False
+    assert user.full_name == user_data["full_name"]
     assert user.Config.orm_mode is True
```
djkresta, line 64: We do not check for orm_mode when creating users previously. Not sure if that will be important.

jobbergateapi2/tests/test_pagination.py
```diff
@@ -19,17 +19,15 @@ def test_pagination_with_default_values():
     """
     pagination = Pagination()
 
-    assert pagination.q is None
     assert pagination.limit == 10
```
djkresta, line 22: Should default hardcoded values be defined at the top of the file or in a config area?

jobbergateapi2/tests/test_pagination.py
```diff
@@ -19,17 +19,15 @@ def test_pagination_with_default_values():
...
 
 def test_pagination_with_custom_values():
     """
     Test the pagination with custom values
     """
-    pagination = Pagination(q="value", limit=100, offset=1)
+    pagination = Pagination(limit=100, skip=1)
 
-    assert pagination.q == "value"
     assert pagination.limit == 100
```
djkresta, line 32: This tests that the limit and the skip are set in the Pagination object, but does not check the functionality. It would be worth checking if the limit actually limits values returned, and that the offset correctly selects the first record to return.

jobbergateapi2/apps/users/routers.py
```diff
@@ -17,35 +17,67 @@
...
+    Return the user given it's id.
+    """
+    query = users_table.select(users_table.c.id == user_id)
+    raw_user = await database.fetch_one(query)
+    if not raw_user:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"User {user_id=} not found",
+        )
+    user = User.parse_obj(raw_user)
```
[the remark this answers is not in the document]
zsinx6: It is a builtin method of the Model to parse the DB object into Python like object

jobbergateapi2/apps/users/routers.py
```diff
@@ -17,35 +17,67 @@
...
         try:
             query = users_table.insert()
             values = {
-                "username": user_data.username,
+                "full_name": user_data.full_name,
                 "email": user_data.email,
                 "password": user_data.hash_password(),
-                "is_admin": user_data.is_admin,
+                "is_superuser": user_data.is_superuser,
+                "is_active": user_data.is_active,
```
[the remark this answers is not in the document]
zsinx6: Sure, I think we can create some exceptions with useful names, but that refactor needs its own PR

jobbergateapi2/main.py
```diff
@@ -3,11 +3,21 @@
 """
 from fastapi import FastAPI
 from mangum import Mangum
+from starlette.middleware.cors import CORSMiddleware
 
 from jobbergateapi2 import storage
+from jobbergateapi2.config import settings
```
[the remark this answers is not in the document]
zsinx6: We take the path from left to right, so `config` < `routers`, so the order is correct. This projects runs some linters to check that everything is correctly formated, so that kind of problem would break the CI

jobbergateapi2/main.py
```diff
@@ -3,11 +3,21 @@
...
 from jobbergateapi2.routers import load_routers
 
 app = FastAPI()
+app.add_middleware(
```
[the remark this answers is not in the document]
zsinx6: I think the CORS is self explained, we are just allowing it in our app

jobbergateapi2/pagination.py
```diff
@@ -8,13 +8,12 @@ class Pagination:
     Basic pagination class
     """
 
-    def __init__(self, q: str = None, limit: int = 10, offset: int = 0):
+    def __init__(self, limit: int = 10, skip: int = 0):
         """
-        Pagination constructor using default limit and offset
+        Pagination constructor using default limit and skip offset.
```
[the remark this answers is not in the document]
zsinx6: When we talk about limit and offset, inside the context of pagination, their meaning must be already clear, so I don't think this add much, but instead it could create some short of confusion to see an unexpected comment for simple things, when nowhere else has them

jobbergateapi2/apps/users/schemas.py
```diff
@@ -19,22 +19,25 @@ class User(BaseModel):
 
     id: Optional[int] = Field(None)
     email: EmailStr
-    is_admin: Optional[bool] = Field(False)
-    username: str = Field(..., max_length=64, description="The name that represents the user")
+    is_active: Optional[bool] = Field(True)
+    is_superuser: Optional[bool] = Field(False)
+    full_name: str = Field(..., max_length=64)
```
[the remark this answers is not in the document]
zsinx6: This can also be added to all fields

jobbergateapi2/tests/apps/users/test_routers.py
zsinx6: Since is_active is default to True and is_superuser is default to False, I think maybe me more clear in the docstring about that stuff. The way you have put it, seems like we are passing the values, when we aren't

jobbergateapi2/tests/apps/users/test_routers.py
zsinx6: It was created, but every test is in its own empty environment, after a test is ran, the database is cleared, that way a test don't interfere with another

jobbergateapi2/tests/apps/users/test_routers.py
```diff
@@ -99,15 +100,56 @@ async def test_create_user_duplication(client, user_data):
...
+    assert response.status_code == status.HTTP_404_NOT_FOUND
+
+
+@pytest.mark.asyncio
+@database.transaction(force_rollback=True)
+async def test_users_me(client, user_data):
+    """
+    Test the /user/me endpoint, should return the current authenticated user.
+    """
+    user = [UserCreate(id=1, **user_data)]
```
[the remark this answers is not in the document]
zsinx6: This is kind of obscure, I get your point, I think this can also be refactored in a next PR alongside with the other fixes.
We are injecting a default token in the `client` and that token has the same email as in the `user_data`, so we just create the user for the token be validated, that same behavior repeats itself for all the other apps

jobbergateapi2/tests/apps/users/test_schemas.py
zsinx6: Each test must be contained in testing just one thing, as a rule of thumb. So in this test we are testing only the schema for the User, so testing it here that the orm_mode is True, is enough, that is already tested here, so we can be sure that always when creating a User schema it has the orm_mode as True

jobbergateapi2/tests/test_pagination.py
zsinx6: Since pagination is a python class, it is defined in the class constructor

jobbergateapi2/tests/test_pagination.py
zsinx6: This test is exactly for that: https://github.com/omnivector-solutions/jobbergate-api-fastapi/blob/a984be82d72ac6d3ad3de876ead9c8aa2c49b251/jobbergateapi2/tests/apps/users/test_routers.py#L38

jobbergateapi2/tests/apps/users/test_routers.py
zsinx6: This is already tested, since len(data) == 0

jamesbeedy (approved):
Really nice work here. No changes requested. 👍

PR #12, opened 2021-05-21
PUT for /users/ to better integrate with frontend

zsinx6:


jobbergateapi2/tests/apps/users/test_routers.py
```diff
@@ -204,3 +204,95 @@ async def test_users_me(client, user_data):
...
+    assert response.status_code == status.HTTP_201_CREATED
+    data = response.json()
+
+    assert data["full_name"] == "new name"
+    assert data["id"] == 1
+
+    query = users_table.select(users_table.c.id == 1)
+    user = User.parse_obj(await database.fetch_one(query))
+
+    assert user is not None
```
djkresta, line 229: is this redundant if we check the full name below this too?

jobbergateapi2/tests/apps/users/test_routers.py
```diff
@@ -204,3 +204,95 @@ async def test_users_me(client, user_data):
...
+    assert response.status_code == status.HTTP_201_CREATED
+    data = response.json()
+
+    assert data["is_superuser"] is False
+    assert data["id"] == 1
+
+    query = users_table.select(users_table.c.id == 1)
+    user = User.parse_obj(await database.fetch_one(query))
+
+    assert user is not None
```
djkresta, line 253: is this redundant if we test is_superuser below?

jobbergateapi2/tests/apps/users/test_routers.py
```diff
@@ -204,3 +204,95 @@ async def test_users_me(client, user_data):
...
+    user = [UserCreate(id=1, **user_data)]
+    await insert_objects(user, users_table)
+
+    response = client.put("/user/me", json={"full_name": "new name"})
+    data = response.json()
+
+    assert data["full_name"] == "new name"
+    assert data["id"] == 1
+
+    assert response.status_code == status.HTTP_201_CREATED
```
djkresta, line 292: should we use HTTP 204 to imply "resource updated successfully"?

jobbergateapi2/tests/apps/users/test_routers.py
zsinx6: It is not, one is checking the return of the request, and the other is checking the data in the database

jobbergateapi2/tests/apps/users/test_routers.py
zsinx6: No, because we are indeed returning a body

jobbergateapi2/tests/apps/users/test_routers.py
zsinx6: It is not, one is checking the return of the request, and the other is checking the data in the database

jobbergateapi2/apps/users/routers.py
```diff
@@ -65,7 +65,61 @@ async def users_me(current_user: User = Depends(get_current_user)):
     return current_user
 
 
-@router.post("/users/", description="Endpoint for user creation", dependencies=[Depends(validate_token)])
+@router.put(
+    "/user/me",
+    description="Endpoint to update current user",
```
jamesbeedy, line 70:
```suggestion
    description="Endpoint to create a user",
```

jamesbeedy (approved):
👍

PR #13, opened 2021-05-27
PUT and DELETE for /job-scripts/

zsinx6:


jobbergateapi2/apps/job_scripts/routers.py
```diff
@@ -219,5 +222,92 @@ async def job_script_list(all: Optional[bool] = Query(None), current_user: User
...
+    description="Endpoint to delete job script",
+)
+async def job_script_delete(
+    current_user: User = Depends(get_current_user),
+    job_script_id: int = Query(..., description="id of the application to delete"),
+):
+    """
+    Delete job_script given its id.
+    """
+    where_stmt = (job_scripts_table.c.id == job_script_id) & (
```
jamesbeedy, line 237: 👍

jobbergateapi2/apps/job_scripts/routers.py
```diff
@@ -219,5 +222,92 @@ async def job_script_list(all: Optional[bool] = Query(None), current_user: User
...
+    job_script_id: int = Query(...),
+    job_script_name: Optional[str] = Form(None),
+    job_script_description: Optional[str] = Form(None),
+    job_script_data_as_string: Optional[str] = Form(None),
+    current_user: User = Depends(get_current_user),
+):
+    """
+    Update a job_script given its id.
+    """
+    query = job_scripts_table.select().where(
```
jamesbeedy, line 269: 👍

jobbergateapi2/apps/job_scripts/routers.py
```diff
@@ -219,5 +222,92 @@ async def job_script_list(all: Optional[bool] = Query(None), current_user: User
...
+
+    values = {
+        "job_script_name": job_script_data.job_script_name,
+        "job_script_description": job_script_data.job_script_description,
+        "job_script_data_as_string": job_script_data.job_script_data_as_string,
+        "updated_at": job_script_data.updated_at,
+    }
+    validated_values = {key: value for key, value in values.items() if value is not None}
+
+    q_update = (
```
jamesbeedy, line 300: 👍

jobbergateapi2/tests/apps/job_scripts/test_routers.py
```diff
@@ -130,6 +131,11 @@ async def test_create_job_script(
 
     count = await database.fetch_all("SELECT COUNT(*) FROM job_scripts")
```
jamesbeedy, line 132: I see these aren't changes in the pr, and I don't really care too much, but in light of being consistent,  why use raw sql here?

zsinx6: It is just easier (and faster too) to get the count, since we are only interested in that

jobbergateapi2/tests/apps/job_scripts/test_routers.py
```diff
@@ -442,3 +448,115 @@ async def test_list_job_script_all(client, user_data, application_data, job_scri
...
+
+    count = await database.fetch_all("SELECT COUNT(*) FROM job_scripts")
+    assert count[0][0] == 1
+
+    response = client.delete("/job-scripts/123")
+
+    assert response.status_code == status.HTTP_404_NOT_FOUND
+
+    count = await database.fetch_all("SELECT COUNT(*) FROM job_scripts")
+    assert count[0][0] == 1
```
zsinx6, line 562: @jamesbeedy here is another good use of the count, since it is way easier and faster to just verify if the actions affected the database

jobbergateapi2/tests/apps/job_scripts/test_routers.py
```diff
@@ -442,3 +448,115 @@ async def test_list_job_script_all(client, user_data, application_data, job_scri
...
+
+    response = client.put("/job-scripts/1", data={"job_script_name": "new name"})
+
+    assert response.status_code == status.HTTP_201_CREATED
+    data = response.json()
+    now = datetime.now()
+
+    assert data["job_script_name"] == "new name"
+    assert data["id"] == 1
+    assert data["updated_at"] == now.isoformat()
```
jamesbeedy, line 477: appropriate use of freeze time feature here - nice.

jobbergateapi2/tests/apps/job_scripts/test_routers.py
```diff
@@ -442,3 +448,115 @@ async def test_list_job_script_all(client, user_data, application_data, job_scri
...
+    assert job_script is not None
+    assert job_script.job_script_name == "new name"
+    assert job_script.updated_at == now
+
+
+@pytest.mark.asyncio
+@database.transaction(force_rollback=True)
+async def test_update_job_script_not_found(client, user_data, application_data, job_script_data):
+    """
+    Test update a job_script not found, must return 404 and do nothing with the stored data.
```
jamesbeedy, line 491: update this to use the long docstring format that we have used in other tests ... "This test proves that ...."

jobbergateapi2/tests/apps/job_scripts/test_routers.py
```diff
@@ -442,3 +448,115 @@ async def test_list_job_script_all(client, user_data, application_data, job_scri
...
+
+    assert job_script is not None
+    assert job_script.job_script_name == job_script_data["job_script_name"]
+
+
+@pytest.mark.asyncio
+@database.transaction(force_rollback=True)
+async def test_delete_job_script(client, user_data, application_data, job_script_data):
+    """
+    Test delete a job_script, must return 204 and delete the data.
```
jamesbeedy, line 517: same docstring update here pls

jobbergateapi2/tests/apps/job_scripts/test_routers.py
```diff
@@ -442,3 +448,115 @@ async def test_list_job_script_all(client, user_data, application_data, job_scri
...
+
+    count = await database.fetch_all("SELECT COUNT(*) FROM job_scripts")
+    assert count[0][0] == 0
+
+
+@pytest.mark.asyncio
+@database.transaction(force_rollback=True)
+async def test_delete_job_script_not_found(client, user_data, application_data, job_script_data):
+    """
+    Test delete a job_script not found, must return 404 and do nothing with the data.
```
jamesbeedy, line 543: same

jobbergateapi2/apps/job_scripts/routers.py
```diff
@@ -219,5 +222,92 @@ async def job_script_list(all: Optional[bool] = Query(None), current_user: User
...
+
+    job_script_data.updated_at = datetime.utcnow()
+
+    values = {
+        "job_script_name": job_script_data.job_script_name,
+        "job_script_description": job_script_data.job_script_description,
+        "job_script_data_as_string": job_script_data.job_script_data_as_string,
+        "updated_at": job_script_data.updated_at,
+    }
+    validated_values = {key: value for key, value in values.items() if value is not None}
```
djkresta, line 298: Why do we need to use a separate values dict? I understand we load our data ( **job_script.dict() ) and then need to compile a dict of all of these key-value pairs that are not None. I believe we could do this in fewer lines without using three dicts, job_script_data, values, and validated values. Something along the lines of:
return {key: value for key, value in JobScript(**job_script.dict()) if value is not None}
Let me know if I am missing something here.

zsinx6: I don't get your suggestion, but we need to make sure the key don't exists in the dict we use for the `values` in the `q_update`, to ensure updating only the changed fields (the ones in the request body). Can you develop more you suggestion?
And we always should prioritize readability instead of fewer lines, and I tried to be as most readable as possible in this function.

jamesbeedy (approved):
LGTM 👍

PR #16, opened 2021-06-09
Permissions for applications

zsinx6:
This apply the fastapi_permissions to the Application resource.
To do that we added the ApplicationPermission resource to manage the permissions for the applications.

The README.md is also updated with a brief explanation on how to use the permission system.

### DIFF
--- a/jobbergateapi2/apps/application_permissions/models.py
+++ b/jobbergateapi2/apps/application_permissions/models.py
@@ -1,0 +1,14 @@
+"""
+Database model for the ApplicationPermission.
+"""
+from sqlalchemy import Integer, String, Table
+from sqlalchemy.sql.schema import Column
+
+from jobbergateapi2.metadata import metadata
+
+application_permissions_table = Table(
+    "application_permissions",
+    metadata,
+    Column("id", Integer, primary_key=True),
+    Column("acl", String, nullable=False, index=True),
+)
--- a/jobbergateapi2/apps/application_permissions/models.py
+++ b/jobbergateapi2/apps/application_permissions/models.py
@@ -10,5 +10,5 @@ application_permissions_table = Table(
     "application_permissions",
     metadata,
     Column("id", Integer, primary_key=True),
-    Column("acl", String, nullable=False, index=True),
+    Column("acl", String, nullable=False, unique=True),
 )
--- a/jobbergateapi2/apps/application_permissions/routers.py
+++ b/jobbergateapi2/apps/application_permissions/routers.py
@@ -1,0 +1,100 @@
+"""
+Router for the ApplicationPermissions resource.
+"""
+import re
+from typing import List
+
+from fastapi import APIRouter, Depends, Form, HTTPException, Query, status
+
+from jobbergateapi2.apps.application_permissions.models import application_permissions_table
+from jobbergateapi2.apps.application_permissions.schemas import ApplicationPermission
+from jobbergateapi2.apps.auth.authentication import get_current_user
+from jobbergateapi2.apps.users.schemas import User
+from jobbergateapi2.compat import INTEGRITY_CHECK_EXCEPTIONS
+from jobbergateapi2.storage import database
+
+router = APIRouter()
+
+
+def check_acl_string(acl_string: str):
+    regex = "^(Allow|Deny)\|[\w:]+\|\w+$"  # noqa
+    if re.match(regex, acl_string):
+        return True
+    else:
+        return False
+
+
+@router.post(
+    "/application-permissions/",
+    status_code=status.HTTP_201_CREATED,
+    description="Endpoint to create application permissions",
+)
+async def application_permissions_create(
+    current_user: User = Depends(get_current_user),
+    acl: str = Form(...),
+):
+    """
+    Create new application permission using an admin user.
+    """
+    if not current_user.is_superuser:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="To create permissions the user must be superuser",
+        )
+
+    if not check_acl_string(acl):
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="Wrong ACL format")
+
+    async with database.transaction():
+        try:
+            query = application_permissions_table.insert()
+            values = {"acl": acl}
+            permission_created_id = await database.execute(query=query, values=values)
+
+        except INTEGRITY_CHECK_EXCEPTIONS as e:
+            raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
+
+    return ApplicationPermission(id=permission_created_id, acl=acl)
+
+
+@router.get(
+    "/application-permissions/",
+    description="Endpoint to list application permissions",
+    response_model=List[ApplicationPermission],
+)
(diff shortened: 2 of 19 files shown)


jobbergateapi2/tests/apps/application_permissions/test_routers.py
```diff
@@ -0,0 +1,247 @@
...
+    Test that is not possible to create an ApplicationPermission with a bad formatted acl.
+
+    This test proves that is not possible to create the ApplicationPermission with a bad formatted acl. We
+    show this by making a request with a bad acl string and asserting that the response status code is 400
+    and that the ApplicationPermission is not created in the database.
+    """
+    user = [UserCreate(is_superuser=True, **user_data)]
+    await insert_objects(user, users_table)
+
+    response = client.post("/application-permissions/", data={"acl": "Allow|"})
```
djkresta, line 115: are there any other bad formats that we should account for? "|view" or "admin|view" or "Allow:|view"

jobbergateapi2/tests/apps/application_permissions/test_routers.py
```diff
@@ -0,0 +1,247 @@
...
+    await insert_objects(user, users_table)
+
+    application_permissions = [
+        ApplicationPermission(id=1, acl="Allow|role:admin|create"),
+        ApplicationPermission(id=2, acl="Allow|role:admin|view"),
+    ]
+    await insert_objects(application_permissions, application_permissions_table)
+
+    response = client.get("/application-permissions/")
+    assert response.status_code == status.HTTP_200_OK
```
djkresta, line 141: maybe change to assert response.status_code == status.HTTP_204_NO_CONTENT

jobbergateapi2/tests/apps/application_permissions/test_routers.py
```diff
@@ -0,0 +1,247 @@
...
+    await insert_objects(user, users_table)
+
+    application_permissions = [
+        ApplicationPermission(id=1, acl="Allow|role:admin|create"),
+        ApplicationPermission(id=2, acl="Allow|role:admin|view"),
+    ]
+    await insert_objects(application_permissions, application_permissions_table)
+
+    response = client.delete("/application-permissions/1")
+    assert response.status_code == status.HTTP_204_NO_CONTENT
```
djkresta, line 192: I believe this should return a 200 success if we are trying to delete 1, and then next we check that 1 is deleted and 2 remains.

djkresta (changes requested):

jobbergateapi2/apps/application_permissions/models.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+from sqlalchemy import Integer, String, Table
+from sqlalchemy.sql.schema import Column
+
+from jobbergateapi2.metadata import metadata
+
+application_permissions_table = Table(
+    "application_permissions",
+    metadata,
+    Column("id", Integer, primary_key=True),
```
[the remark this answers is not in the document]
zsinx6: It is very common to use `id` for database id and it doesn't will be a problem in any moment because it is a "property" of some object. Actually it is better to use id here, than some other, since the user expects the table to have a column id.

jobbergateapi2/apps/application_permissions/routers.py
```diff
@@ -0,0 +1,99 @@
...
+    where_stmt = application_permissions_table.c.id == permission_id
+    get_query = application_permissions_table.select().where(where_stmt)
+    raw_permission = await database.fetch_one(get_query)
+    if not raw_permission:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"ApplicationPermission {permission_id=} not found.",
+        )
+    delete_query = application_permissions_table.delete().where(where_stmt)
+    await database.execute(delete_query)
```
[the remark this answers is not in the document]
zsinx6: It is not necessary, since we already know that the object exists in the database, if the database is well everything will work. In this case is better to let it break if something is wrong with the database, so we can take action on that.

jobbergateapi2/apps/applications/routers.py
```diff
@@ -48,7 +74,7 @@ async def applications_create(
             application_created_id = await database.execute(query=query, values=values)
 
         except INTEGRITY_CHECK_EXCEPTIONS as e:
-            raise HTTPException(status_code=422, detail=str(e))
+            raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
     application_location = f"{settings.S3_BASE_PATH}/{application.application_owner_id}/applications/{application_created_id}/jobbergate.tar.gz"  # noqa
```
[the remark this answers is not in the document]
zsinx6: It is indeed, but since it is the line to configure the location of the application, it will be ugly if we break that. There is already a noqa in the line to indicate that to the linter. It would be a problem if we have a lot of them, but there is only here.

jobbergateapi2/apps/applications/routers.py
```diff
@@ -48,7 +74,7 @@ async def applications_create(
             application_created_id = await database.execute(query=query, values=values)
 
         except INTEGRITY_CHECK_EXCEPTIONS as e:
-            raise HTTPException(status_code=422, detail=str(e))
+            raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
```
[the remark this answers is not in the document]
zsinx6: Yes, we only have unique constraints for that table, it is the same case as the previous question in case of another kind of problem.

jobbergateapi2/apps/auth/authentication.py
```diff
@@ -38,16 +39,31 @@ def validate_token(token: str = Depends(oauth2_scheme)):
...
     user = User.parse_obj(await database.fetch_one(query))
     return user
 
 
+def get_active_principals(user: User = Depends(get_current_user)):
+    """
+    Return the principals for the active user.
+    """
+    if user:
+        principals = user.principals.split("|") if user.principals != "" else []
```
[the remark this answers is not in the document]
zsinx6: It is pretty straightforward, I think the comment will only confuse the reader, the ternary operator is very common in python, and here the conditional is very simple too

jobbergateapi2/tests/apps/application_permissions/test_routers.py
zsinx6: Yes, but here we don't need to test them all, since there is already a test for that in the test_schemas for the application_permissions.

jobbergateapi2/tests/apps/application_permissions/test_routers.py
zsinx6: Since it is ok to return 204 for GET, in this case I think it only adds unnecessary complexity to the code, and since it returned a empty list, it is not `no content` because the the list is there.

jobbergateapi2/tests/apps/application_permissions/test_routers.py
zsinx6: About the status code, here we return 204  since there is no body in the response, 200 would fit if we were expecting something if the response body, which is not the case.

jamesbeedy:
👍  LGTM - Very nice work here!

djkresta (approved):
Looks good!

matheushent (approved):

PR #17, opened 2021-06-15
Permissions refactor to allow multiple resources

zsinx6:
In order to reuse the code, I made a major refactor to how the permissions are implemented, now it is necessary to send the permission type in the query string:
`/permissions/?application`, this is validated by a regex in the `jobbergateapi2.apps.permissions.routers._QUERY_RX`
And now all the tests are also refactored to allow testing all the resources at once.


jamesbeedy (approved):
💯 👍 

Excellent work on this. Thank you!

PR #25, opened 2021-07-15
Add prestart.sh to create the first user

zsinx6:
#### What
Add prestart.sh script.

#### Why
To create the first user using email and password provided as ENV_VAR.

For example:
`docker run -d -p 80:80 --name jobbergateapi2 -e SECRET_KEY="changeme" -e EMAIL="admin@email.com" -e PASSWORD="123" jobbergateapi2:latest`

`Task`: https://omnivector-solutions.monday.com/boards/1206528980/pulses/1473686557

### DIFF
--- a/prestart.sh
+++ b/prestart.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+sleep 10;
+createsuperuser --email=$EMAIL --password=$PASSWORD --full-name=admin
--- a/prestart.sh
+++ b/prestart.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-sleep 10;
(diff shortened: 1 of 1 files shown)


prestart.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+sleep 10;
+createsuperuser --email=$EMAIL --password=$PASSWORD --full-name=admin
```
heitorPB, line 4:
```suggestion
sleep 10
createsuperuser --email="$EMAIL" --password="$PASSWORD" --full-name=admin
```

Using `;` is not necessary at end of lines in bash

jamesbeedy: We should also use `${foo}` for good practice as noted here https://stackoverflow.com/a/8748880

heitorPB (approved):
LGTM

PR #27, opened 2021-08-11
Fix staging

zsinx6:
#### What
Fix CORS and createsuperuser script.

#### Why
To be able to configure CORS using env in terraform.

### DIFF
--- a/jobbergateapi2/config.py
+++ b/jobbergateapi2/config.py
@@ -16,6 +16,7 @@ class Settings(BaseSettings):
     SERVERLESS_STAGE: str = Field("staging")
     SERVERLESS_REGION: str = Field("eu-north-1")
     S3_BASE_PATH: str = Field("jobbergate-resources")
+    # BACKEND_CORS_ORIGINS example: "['https://example1.com', 'https://example2.com']"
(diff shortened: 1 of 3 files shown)


jobbergateapi2/config.py
```diff
@@ -18,7 +18,7 @@ class Settings(BaseSettings):
     SERVERLESS_STAGE: str = Field("staging")
     SERVERLESS_REGION: str = Field("eu-north-1")
     S3_BASE_PATH: str = Field("jobbergate-resources")
-    BACKEND_CORS_ORIGINS: List[AnyHttpUrl] = []
+    BACKEND_CORS_ORIGINS: str = Field("[]")
```
matheushent, line 21: what is the reason for changing it?

zsinx6: Take a look here: https://github.com/omnivector-solutions/infrastructure/pull/18

matheushent: can you add an example in anywhere? e.g. `"['https://example1.com', 'https://example2.com']"`. It seems dummy but can save some understanding in the future

jamesbeedy (approved):
Nice. +1

PR #28, opened 2021-08-25
[Draft] Tucker/migrate to armada security for auth

dusktreader:
#### What
This change updates Jobbergate so that it can use security tokens issued by auth0. It makes heavy use of the armada-security package to allow securing of the endpoints with very little code.

#### Why
We need to share the same users and permissions across the various apps. We also want to use a 3rd party auth and user administration tool so we don't have to build it all ourselves.

`Task`: https://omnivector-solutions.monday.com/boards/1206528980/pulses/1430260172

---

#### Peer Review


.github/workflows/test_on_push.yaml
```diff
@@ -7,23 +7,15 @@ on:
...
-    name: "Test code and check lint"
+  tests:
+    name: "Tests"
     runs-on: "ubuntu-latest"
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.x'
+          python-version: '3.8'
```
zsinx6, line 17: Is there a reason to use 3.8?

dusktreader: Nope, not really. Will update

jamesbeedy: We should create/document a dev workflow that provides python at the specified version since none of us will have the same system python when deving locally. If we are going to be pinning python in the test/build environment then we should also provide a way for that same python to be used in the local setup.

.gitignore
```diff
@@ -108,3 +108,6 @@ venv.bak/
 
 # sqlite
 *.db
```
jamesbeedy, line 110: If we are dropping support for sqlite then we can probably remove these lines from the .gitignore

.gitignore
```diff
@@ -108,3 +108,6 @@ venv.bak/
 
 # sqlite
```
jamesbeedy, line 109: If we are dropping support for sqlite then we can probably remove these lines from the .gitignore

.gitignore
```diff
@@ -108,3 +108,6 @@ venv.bak/
...
 *.db
+
```
jamesbeedy, line 111: If we are dropping support for sqlite then we can probably remove these lines from the `.gitignore`.

jamesbeedy, line 111: If we are dropping support for sqlite then we can probably remove these lines from the `.gitignore`.

Dockerfile
```diff
@@ -1,8 +1,24 @@
...
+RUN mkdir app
+WORKDIR app
+
+COPY ./pyproject.toml /app/
+COPY ./poetry.lock /app/
+
+COPY hack_build_secrets.sh .
+RUN --mount=type=secret,id=pypi_password ./hack_build_secrets.sh
+ENV POETRY_HTTP_BASIC_PYPICLOUD_USERNAME=admin
+ENV POETRY_HTTP_BASIC_PYPICLOUD_PASSWORD=$PYPI_PASSWORD
```
jamesbeedy, line 19: +1

Dockerfile-dev
```diff
@@ -0,0 +1,26 @@
...
+RUN pip install 'poetry==1.1.7'
+
+ENV MODULE_NAME="jobbergateapi2.main"
+
+COPY ./pyproject.toml /app/
+COPY ./poetry.lock /app/
+
+ARG PYPI_PASSWORD
+ENV POETRY_HTTP_BASIC_PYPICLOUD_USERNAME=admin
+ENV POETRY_HTTP_BASIC_PYPICLOUD_PASSWORD=$PYPI_PASSWORD
```
jamesbeedy, line 15: why not use `hack_build_secrets.sh` here too?

Dockerfile-dev
```diff
@@ -0,0 +1,26 @@
...
+
+ARG AWS_ACCESS_KEY_ID
+ENV AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID
+
+ARG AWS_SECRET_ACCESS_KEY
+ENV AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
```
jamesbeedy, line 21: should these be added by `hack_build_secrets.sh` too? Or is it just easier in local dev to pull the stuff in from args?

jobbergateapi2/apps/applications/routers.py
```diff
@@ -61,11 +52,11 @@ async def applications_create(
 
         except INTEGRITY_CHECK_EXCEPTIONS as e:
             raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
-    application_location = f"{settings.S3_BASE_PATH}/{application.application_owner_id}/applications/{application_created_id}/jobbergate.tar.gz"  # noqa
+    application_location = "{base}/{owner_id}/applications/{app_id}/jobbergate.tar.gz".format(
+        base=settings.S3_BASE_PATH, owner_id=application.application_owner_id, app_id=application_created_id,
+    )
```
jamesbeedy, line 57: Much more clear/maintainable this way, +1

jobbergateapi2/apps/applications/routers.py
```diff
@@ -89,36 +79,33 @@ async def application_delete(
...
-            detail=f"Application {application_id=} not found.",
+            status_code=status.HTTP_404_NOT_FOUND, detail=f"Application {application_id=} not found.",
         )
     application = Application.parse_obj(raw_application)
     delete_query = applications_table.delete().where(where_stmt)
     await database.execute(delete_query)
-    application_location = f"{settings.S3_BASE_PATH}/{application.application_owner_id}/applications/{application.id}/jobbergate.tar.gz"  # noqa
+    application_location = "{base}/{owner_id}/applications/{app_id}/jobbergate.tar.gz".format(
+        base=settings.S3_BASE_PATH, owner_id=application.application_owner_id, app_id=application_id,
+    )
```
jamesbeedy, line 89: Can we make a little helper out of this since we use it in > 1 place?

jobbergateapi2/apps/applications/routers.py
```diff
@@ -197,17 +180,17 @@ async def application_update(
 
         except INTEGRITY_CHECK_EXCEPTIONS as e:
             raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
-    application_location = (
-        f"{settings.S3_BASE_PATH}/{current_user.id}/applications/{application_id}/jobbergate.tar.gz"
+    query = applications_table.select().where(applications_table.c.id == application_id)
+    application = Application.parse_obj(await database.fetch_one(query))
+
+    application_location = "{base}/{owner_id}/applications/{app_id}/jobbergate.tar.gz".format(
+        base=settings.S3_BASE_PATH, owner_id=application.application_owner_id, app_id=application_id,
```
jamesbeedy, line 187: same ^^

jobbergateapi2/apps/job_submissions/schemas.py
```diff
@@ -19,12 +19,7 @@ class JobSubmissionRequest(BaseModel):
 
     class Config:
         orm_mode = True
-        schema_extra = {
-            "example": {
-                "job_submission_name": "name",
-                "job_script_id": 1,
-            }
-        }
+        schema_extra = {"example": {"job_submission_name": "name", "job_script_id": 1}}
```
jamesbeedy, line 22: Why do we have this `schema_extra`? A comment here about what this does would be helpful.

jobbergateapi2/config.py
```diff
@@ -1,24 +1,29 @@
...
 
 
 class Settings(BaseSettings):
     DATABASE_URL: str = Field("sqlite:///./sqlite.db?check_same_thread=true", regex=_DB_RX)
     TEST_ENV: bool = Field(False)
-    SECRET_KEY: str
-    TOKEN_URL: str = Field("token/")
-    ALGORITHM: str = Field("HS256")
-    ACCESS_TOKEN_EXPIRE_MINUTES: int = Field("30")
     SERVERLESS_STAGE: str = Field("staging")
```
jamesbeedy, line 12: I think these `SERVERLESS_*` can be removed as they were only used for serverless, and we don't use serverless anymore.

jamesbeedy (changes requested):
Really nice work here. I left some comments to resolve and I would like to see module level docstrings. Other than that, +1.

jamesbeedy (changes requested):
Really nice work here. I left some comments to resolve and I would like to see module level docstrings. Other than that, +1.

.github/workflows/test_on_push.yaml
dusktreader: I think pyenv is the best tool for this. I can write a quick walkthrough, but where should that go? In our osd-documentation repo?

dusktreader: Added this pull-request to address that. Please let me know what you think. https://github.com/omnivector-solutions/Documentation/pull/12

dusktreader: Addressed the original comment in the latest commit

.gitignore
dusktreader: This repo still supports sqlite for testing.

Dockerfile-dev
dusktreader: Because this Dockerfile-dev is intended to be used for local development (with docker-compose), and secrets will probably not be stored on your workstation in the same way.

To use this Dockerfile-dev by itself, you would do this:
```bash
$ docker build --file=Dockerfile-dev --build-arg="PYPI_PASSWORD=putpasswordhere" .
```

dusktreader: I'm still working on setting up the armada-composed `docker-compose.yaml` file, but I will include in the README a description of how to set the build args.

Dockerfile-dev
dusktreader: You are absolutely right

dusktreader: Well, in the prod container, it should be pulling AWS stuff from docker secrets. Unlike the PYPI_PASSWORD, these aren't needed at container build time.

For local build it seemed easy at the moment to just include them as a build-arg.

jobbergateapi2/apps/applications/routers.py
dusktreader: Created a manager class for the s3 client that wraps all the s3 interactions up in a nice little interface. See latest commit...

jobbergateapi2/apps/job_submissions/schemas.py
dusktreader: I had to look this up myself. This just adds an example in the swagger pages. I added a comment with the link.

dusktreader:
Need to add documentation about configuring Armasec via env variables

jobbergateapi2/tests/apps/job_scripts/conftest.py
```diff
@@ -1,3 +1,6 @@
+"""
+Add convenience fixtures for the job scripts tests.
```
jamesbeedy, line 2:
```suggestion
Convenience fixtures for the job scripts tests.
```

jobbergateapi2/tests/apps/job_submissions/conftest.py
```diff
@@ -1,3 +1,6 @@
+"""
+Adds job_submission_data fixture for job submission tests
```
jamesbeedy, line 2:
```suggestion
Job_submission_data fixture for job submission tests
```

bin/db_login.py
```diff
@@ -0,0 +1,7 @@
+import subprocess
+
+from jobbergateapi2.config import settings
+
+
+def login():
+    subprocess.run(["pgcli", settings.DATABASE_URL])
```
zsinx6, line 7: What is this file for?

dusktreader: This is a convenience script for logging into the database that the project is set up to use. Once you have your configuration set up, you can simply run:
```
$ poetry run db-login
```

And you will be dropped into an interactive database shell.

It saves you from having to remember and type in the full postgres command every time.

PR #33, opened 2021-11-10
Tucker/job 82  migrate jobbergate api

dusktreader:
#### What
Migrated the jobbergate-cli into the jobbergate monorepo

#### Why
See: https://app.clickup.com/t/18022949/JOB-80

---

#### Peer Review


jobbergate-cli/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+.PHONY: test
+test: install
+	poetry run pytest
+
+.PHONY: lint
+lint: install
+	poetry run black --check ${PACKAGE_NAME}
+	poetry run isort --check ${PACKAGE_NAME}
+	poetry run flake8 --max-line-length=120 --max-complexity=40 ${PACKAGE_NAME}
```
matheushent, line 20: what about having a .flake8 file here? Example: https://github.com/omnivector-solutions/armada-api/blob/main/.flake8

dusktreader: I prefer not to have an extra configuration file. I wish flake8 supported pyproject.toml like the other tools we use, but it won't (https://github.com/PyCQA/flake8/issues/234#issuecomment-812800832). Instead of adding another config file just for flake8, I prefer to specify the config in the Makefile.

jobbergate-cli/pyproject.toml
```diff
@@ -0,0 +1,56 @@
...
+pytest-random-order = "^1.0.4"
+pytest-freezegun = "^0.4.2"
+pytest-responsemock = "^1.0.1"
+flake8 = "^3.9.2"
+isort = "^5.9.3"
+pytest-env = "^0.6.2"
+
+[tool.pytest.ini_options]
+minversion = "6.0"
+addopts = "--random-order --cov=jobbergate_cli --cov-report=term-missing --cov-fail-under=25"
```
matheushent, line 36: can you do this as list like [here](https://github.com/omnivector-solutions/armada-api/blob/main/pyproject.toml#L74-L82)? I think it's better to see

jobbergate-cli/pyproject.toml
```diff
@@ -0,0 +1,56 @@
...
+env = [
+    "JOBBERGATE_API_ENDPOINT = https://jobbergate-api-staging.omnivector.solutions",
+    "JOBBERGATE_DEBUG = false",
+]
+testpaths = ["jobbergate_cli/test"]
+
+[tool.black]
+
```
matheushent, line 44: shouldn't `line-length` be set with value 120?

dusktreader: There was some discussion about this at one point. Folks were going back and forth between 100 and 120, so I split the difference. It's consistent across black, flake8, and isort in this project.

matheushent: I just mean because flake8 is set to check 120 but the default is 110. I'm just used to keep it the same

jobbergate-cli/pyproject.toml
matheushent: I think you changed another pyproject.toml. This one remains the same

dusktreader: Oh, I changed in jobbergate-api. I'll update it in jobbergate-cli as well.

jobbergate-cli/pyproject.toml
dusktreader: In the latest commit, I standardized the max line-length to 110 on both jobbergate-api and jobbergate-cli

matheushent (approved):
LGTM

PR #38, opened 2021-11-18
Tucker/job 71  changelog but in new sub project

dusktreader:
#### What
Added a CHANGELOG.

Also pinned project version to 0.0.1 until first release. This will allow us to import the release script from jobbergate-cli and have the first proper release be versioned 0.1.0

#### Why
We needed a CHANGELOG

`Task`: https://app.clickup.com/t/18022949/JOB-71

---

#### Peer Review

### DIFF
--- a/jobbergate-api/pyproject.toml
+++ b/jobbergate-api/pyproject.toml
@@ -5,7 +5,7 @@ authors = [
 name = "jobbergateapi2"
 description = "jobbergate-api"
 license = "MIT"
-version = "0.0.1"
+version = "0.1.0"
 
 [tool.poetry.dependencies]
 python = "^3.8"
--- a/jobbergate-api/pyproject.toml
+++ b/jobbergate-api/pyproject.toml
@@ -5,7 +5,7 @@ authors = [
 name = "jobbergateapi2"
 description = "jobbergate-api"
 license = "MIT"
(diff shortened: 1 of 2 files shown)


jobbergate-api/pyproject.toml
```diff
@@ -5,7 +5,7 @@ authors = [
 name = "jobbergateapi2"
 description = "jobbergate-api"
 license = "MIT"
-version = "0.1.0"
+version = "0.0.1"
```
heitorPB, line 8: ??????

dusktreader: I did this so the first time we run the release script, we can bump the minor version. It's still unreleased, so this doesn't mean that much at this point. Poetry sets the initial version to 0.1.0 by default, but I want that to be the first version we actually release.

If it's too weird, I can just make 0.2.0 the first released version

heitorPB: Let's not roll back versions, this can cause a lot of confusion.

dusktreader: It's unreleased. It's not really a rollback.

heitorPB (approved):

jamesbeedy (approved):
👍

PR #39, opened 2021-11-19
Tucker/job 86  add auth0 token support to jobbergate cli

dusktreader:
#### Update Summary
    This change allows a user to authenticate requests against the
    jobbergate-api using a token retrieved from the Armada frontend.

    The token is added by using the --token option.

    Also:
    * Updated docker-compose with network for bridged communication (to Auth0)
    * Moved lifecycle events in jobbergate-api to main app
    * Converted jobbergate-cli to use Pydantic for settings
    * Moved constants to a constants module
    * Removed unused config/*.json files
    * Converted broken urljoin in api wrapper to use yarl instead
    * Captured stack traces for errors in jobbergate-cli logs
    * Allowed CLI to be invoked without subcommand (for just binding token)
    * Updated unit tests
    * Updated dependencies

`Task`: https://app.clickup.com/t/18022949/JOB-86

---

#### Peer Review


jobbergate-api/jobbergateapi2/main.py
```diff
@@ -43,14 +48,10 @@ async def init_database():
...
 
 
-@subapp.on_event("shutdown")
+@app.on_event("shutdown")
 async def disconnect_database():
     """
     Disconnect the database
     """
     logger.debug("Disconnecting database")
     await storage.database.disconnect()
```
matheushent, line 57: there is a specific PR for this change: https://github.com/omnivector-solutions/jobbergate/pull/40. Should we close that one or rollback here?

matheushent: I think it's better to rollback here since the PR is specific for the cli feature, right?

jobbergate-cli/jobbergate_cli/config.py
```diff
@@ -0,0 +1,70 @@
...
+        values["JOBBERGATE_USER_TOKEN_DIR"] = token_dir
+        values["JOBBERGATE_API_JWT_PATH"] = token_dir / "jobbergate.token"
+
+        return values
+
+    class Config:
+        if constants.JOBBERGATE_DEFAULT_DOTENV_PATH.is_file():
+            env_file = constants.JOBBERGATE_DEFAULT_DOTENV_PATH
+        else:
+            env_file = ".env"
```
matheushent, line 67: shouldn't we use a specific dotenv file like many other apps do? e.g. ~/.aws/credentials, ~/.kube/config, etc

jobbergate-cli/jobbergate_cli/constants.py
```diff
@@ -0,0 +1,35 @@
...
+    "job_submission_name": "",
+    "job_submission_description": "TEST_DESC",
+    "job_submission_owner": "",
+    "job_script": "",
+}
+
+JOBBERGATE_DEFAULT_DOTENV_PATH = Path("/etc/default/jobbergate-cli")
+JOBBERGATE_APPLICATION_MODULE_FILE_NAME = "jobbergate.py"
+JOBBERGATE_APPLICATION_CONFIG_FILE_NAME = "jobbergate.yaml"
+TAR_NAME = "jobbergate.tar.gz"
```
matheushent, line 31: I think each file should be named diferrently, otherwise, the files can be overwritten. What about something like: `<user-id>-<datetime>.tar.gz`?

matheushent (changes requested):

jobbergate-cli/jobbergate_cli/constants.py
matheushent: the user_id in this case can be the same one used for Sentry. Thoughts?

jobbergate-api/jobbergateapi2/main.py
dusktreader: yes, will do that.

jobbergate-cli/jobbergate_cli/config.py
dusktreader: We do for the default (/etc/default/jobbergate-cli/.env). However, we want to be able to use one that is in the project directory for local development.

jobbergate-cli/jobbergate_cli/constants.py
dusktreader: The files don't collide because the application id is included in the path.

See the `key_template` attribute in jobbergateapi2/s3_manager.py.

jobbergate-api/jobbergateapi2/main.py
dusktreader: revised in latest commit.

jobbergate-cli/jobbergate_cli/constants.py
matheushent: even so, different users can face different problems with the same application. Identifying the user is useful I guess. Thoughts?

dusktreader: The ownership of the application is bound in the database. We shouldn't include it in the path on S3. We were doing this before, and it caused problems because if one user updated an application owned by a different user, it would create a new entry with a different user id but the same application id. Since the application id is unique to an instance of the application, we should include only that in the application's s3 path. If there are issues with an application (indicated by an error log), we should consult the database if we need to learn who the owner is. Beyond that, the logs are collected user-by-user so there will be breadcrumbs there to indicate who had the issue.

jamesbeedy: I agree, there is no need to make a unique filename here.

jamesbeedy: I see your point @matheushent , but it adds unneeded complexity. The value of creating a named file here will only serve developers when they look in s3, it will not add value to the code as far as I can see. Resolving this issue so we can move forward.

matheushent (approved):

PR #62, opened 2022-01-20
JOB-120 Added  param to job-submissions list endpoint

dusktreader:
What
Added a new optional parameter to the list job-submissions endpoint.

Why
We need a way to get all job_submissions that have a matching slurm job id (this is needed for the frontend).

Task: https://app.clickup.com/t/18022949/JOB-120

---

#### Peer Review

### DIFF
--- a/jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
+++ b/jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
@@ -92,8 +92,7 @@ async def job_submission_list(
         None, description="If supplied, do not limit job_submissions to only the current user",
     ),
     slurm_job_ids: Optional[str] = Query(
-        None,
-        description="Non-empty Comma-separated list of slurm-job-ids to match active job_submissions",
+        None, description="Comma-separated list of slurm-job-ids to match active job_submissions",
     ),
     token_payload: TokenPayload = Depends(guard.lockdown(Permissions.JOB_SUBMISSIONS_VIEW)),
 ):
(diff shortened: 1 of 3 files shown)


jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -87,16 +87,35 @@ async def job_submission_get(job_submission_id: int = Query(...)):
...
+
+    :param: all:           Do not limit results to only the current user
+    :param: slurm_job_ids: Limit results job_submissions having one of the supplied slurm_job_ids.
+                           If used, must be a comma separated list of integers.
     """
     identity_claims = IdentityClaims.from_token_payload(token_payload)
     query = job_submissions_table.select()
     if not all:
         query = query.where(job_submissions_table.c.job_submission_owner_email == identity_claims.user_email)
+    if slurm_job_ids is not None:
```
heitorPB, line 110: If the user sends an empty list, will it work also?

dusktreader: Good point. Let me add some logic and unit-tests for that.

dusktreader: @heitorPB Addressed in the latest commit.

heitorPB: I think an empy list should have the same behaviour as not having the list. Why you decided to return an error to the user?

dusktreader: The issue is that we already have a null case for this (when the value is `None`). If we are saying in the docs that the argument should be a comma separated list of ids, then empty string to me seems invalid even if it would work functionally. I made the error condition to remove some ambiguity between `""` and `None`.

heitorPB: Slurm does not make this distinction between empty and not defined. I think Jobbergate should follow this strategy. Otherwise it can be confusing to end users.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -87,16 +87,39 @@ async def job_submission_get(job_submission_id: int = Query(...)):
...
+            job_ids = [int(i) for i in slurm_job_ids.split(",")]
+            if len(job_ids) == 0:
+                raise ValueError
+        except Exception:
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail="Invalid slurm_job_ids param. Must be a non-empty comma-separated list of ints",
+            )
+
+        query = job_submissions_table.select().where(job_submissions_table.c.slurm_job_id.in_(job_ids))
```
heitorPB, line 122: I don't get the logic of this function. The first `query` statement will be overwritten if the user sends a list of jobs?

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: I feel like we're doing a bit of bike-shedding here.

I'm going to remove the extra logic in the implementation from the last commit but still include a unit test to show that empty string has no effect on the filter.

dusktreader: Resolved in latest commit.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: Thank you, that was a bad copy-pasta error. It should hav ebeen applying the `where()` to the previous query. This has been corrected in the latest commit.

jamesbeedy (approved):
👍

PR #73, opened 2022-02-10
Fixed message for deletes due to foreign key errors

dusktreader:
#### What
Provided better, more helpful error messages when a delete fails due to a foreign key constraint

#### Why
Because we need to guide the user better when something doesn't work

---

#### Peer Review

### DIFF
--- a/jobbergate-cli/jobbergate_cli/jobbergate_api_wrapper.py
+++ b/jobbergate-cli/jobbergate_cli/jobbergate_api_wrapper.py
@@ -173,6 +173,18 @@ class JobbergateApi:
                     solution="Please confirm the id and try again",
                 )
                 return response
+            elif response.status_code == 409:
+                try:
+                    body = response.json()
+                    error = body["detail"]["message"]
+                    table = body["detail"]["table"]
+                    pk_id = body["detail"]["pk_id"]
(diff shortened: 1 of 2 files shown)


jobbergate-cli/jobbergate_cli/jobbergate_api_wrapper.py
```diff
@@ -173,6 +173,17 @@ def jobbergate_request(self, method, endpoint, data=None, files=None, params=Non
...
                 return response
+            elif response.status_code == 409:
+                try:
+                    body = response.json()
+                    error = body["detail"]["message"]
+                    table = body["detail"]["table"]
+                    solution = f"First delete all items from {table} that depend on the item to be deleted"
+                except:
+                    error = "There was a conflict encountered when deleting"
+                    solution = "Please alert Omnivector for resolution"
```
matheushent, line 184: I wonder how much helpful this is. Can we provide details to the user? That way the user can come to use with more information and we can be more precise.

dusktreader: This is a fallback for when we don't know what the issue was in the API. It shouldn't happen, but this is here as a contingency. The information/details are collected in the logs and sentry events.

matheushent (approved):

PR #79, opened 2022-02-14
Fixed parameters to api for create job scripts and submissions

dusktreader:
#### What
Adjusted api call parameters

#### Why
The CLI was supplying "application" instead of "application_id", and the API wasn't happy about it.


jamesbeedy (approved):
👍

PR #83, opened 2022-02-16
Fixed urls for client credentials login

dusktreader:
#### What
Added an AUTH0_LOGIN_DOMAIN setting to use for requests from jobbergate-cli

#### Why
We need to be able to use login.omnivector.solutions for login. This will probably change in the future to a more armada specific domain, so I made it an env setting that can be adjusted without a code change.

---

#### Peer Review


jamesbeedy (approved):
Niceee 👍

PR #84, opened 2022-03-01
JOB-93 Refactored Jobbergate API to use postgres for unit tests

dusktreader:
`Task`: https://app.clickup.com/t/18022949/JOB-93

Changed the database backend to postgres. This requires a database
instance to be running externally. This may be done by standing up the
testing database via `docker-compose` or by using the new `jobbergate db
start-test` command. The test performance is slower, but the validation
is improved because the tests better represent the environmen in which
the code will ultimately be running.

Notable changes:
* Added some cli helpers based on `py-docker-gadgets` to stand up
  database instances with docker
* Moved command-line helpers in to a `cli` submodule and rebuilt them as
  a single typer app with sub-commands
* Removed `DATABASE_URL` as a configruation option. Instead, the url
  MUST always be built from parts
* Added `TEST_DATABASE_` configuration settings to make sure that the
  tests cannot accidently target a non-test database
* Added `TEST` item to DeployEnvEnum. This is explicitly set in the
  pytest env section of pyproject.toml to ensure tests target test DBs.
* Updated the docker-compose to improve usability in a dev environment
* Added pyproject-flake8 to allow flake8 config to live in
  pyproject.toml
* Used `returning` clauses for `create` and `update` endpoints to avoid
  a second query. This is only possible in postgres (SQLite doesn't
  support `returning`)
* Removed double-serialization of data responses. FastAPI does
  serialization automatically; no need to do it explicitly in the methods
* Removed explicit `with database.transaction` clauses. Databases
  already creates transactions for async tasks automatically.
* Added an error handler for handling foreign key errors so that the
  `handle_fk_error` context manager can instead just be used globally.
* Removed `compat` module: it's not necesary any more.
* Removed `create_all()` from main. Non-test databases should use
  alembic to apply schema changes.
* Reworked pagination/packaging for list responses. Added an
  `ok_response` helper method to make response typing on list endpoints
  re-usable.
* Added `fill_.*_data` fixtures to make test setup easier (combine
  default data with explicitly set fields for the test cases)
* Removed reliance in unit tests on primary key values and removed
  assignment of primary keys. This does not work in postgres due to
  persistent pk-indexes even after rollback.
* Adjusted test fixtures that dealt with the database to be more
  correct.
* Lots of fixes and miscellaneous adjustments to unit tests.
* Bumped IPython dependency due to security issue. This was not severe
  since IPyhton is only a dev dependency, but it will shut dependabot up

---

#### Peer Review


docker-compose.yml
```diff
@@ -11,38 +11,51 @@ services:
...
-      - DATABASE_HOST=${NEXTGEN_DATABASE_HOST:-nextgen-db}
-      - DATABASE_USER=${NEXTGEN_DATABASE_USER:-nextgen}
-      - DATABASE_PSWD=${NEXTGEN_DATABASE_PSWD:-nextgen}
-      - DATABASE_NAME=${NEXTGEN_DATABASE_NAME:-nextgen}
-      - DATABASE_PORT=${NEXTGEN_DATABASE_PORT:-5432}
-      - S3_ENDPOINT_URL=${S3_ENDPOINT_URL:-http://s3:9000}
+      - DATABASE_HOST=${DATABASE_HOST:-db}
+      - DATABASE_USER=${DATABASE_USER:-omnivector}
+      - DATABASE_PSWD=${DATABASE_PSWD:-local-pswd}
+      - DATABASE_NAME=${DATABASE_NAME:-jobergate}
```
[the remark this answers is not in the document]
dusktreader: Fixed in latest commit.

docker-compose.yml
```diff
@@ -11,38 +11,51 @@ services:
...
       - jobbergate-net
     volumes:
       - postgres_data:/var/lib/postgresql/data/
     environment:
-      - POSTGRES_PASSWORD=${NEXTGEN_DATABASE_PSWD:-nextgen}
-      - POSTGRES_USER=${NEXTGEN_DATABASE_USER:-nextgen}
-      - POSTGRES_DB=${NEXTGEN_DATABASE_NAME:-nextgen}
+      - POSTGRES_PASSWORD=${NEXTGEN_DATABASE_PSWD:-local-pswd}
+      - POSTGRES_USER=${NEXTGEN_DATABASE_USER:-omnivector}
+      - POSTGRES_DB=${NEXTGEN_DATABASE_NAME:-jobbergate}
```
matheushent, line 41: not sure if this is useful. Can't you just set dummy fixed params here?

jobbergate-api/jobbergate_api/cli/db.py
```diff
@@ -0,0 +1,55 @@
...
+        ),
+        ports={"5432/tcp": settings.TEST_DATABASE_PORT},
+        tmpfs={"/var/lib/postgresql/data": "rw"},
+    )
+
+
+@app.command()
+def start_all():
+    start_dev()
+    start_test()
```
matheushent, line 55: is this CLI app really useful? I just see complexity being added to the project and more code to maintain in the long run, don't you?

jobbergate-api/jobbergate_api/cli/dev_server.py
```diff
@@ -0,0 +1,42 @@
...
+    db_wait_count: int = typer.Option(3, help="How many times to attempt a check"),
+    db_wait_interval: float = typer.Option(5.0, help="Seconds to wait between checks"),
+    port: int = typer.Option(5000, help="The port where the server should listen"),
+    log_level: str = typer.Option("DEBUG", help="The level to log uvicorn output"),
+):
+    try:
+        _wait_for_db(db_wait_count, db_wait_interval)
+    except Exception:
+        logger.error("Database is not available")
+        typer.Exit(1)
```
matheushent, line 38: a good docker-compose.yaml file wouldn't require this. Take the [armada API example](https://github.com/omnivector-solutions/armada-api/blob/main/docker-compose.yaml#L26-L38)

jobbergate-api/jobbergate_api/config.py
```diff
@@ -32,13 +32,19 @@ class Settings(BaseSettings):
...
-    DATABASE_PSWD: Optional[str]
-    DATABASE_NAME: Optional[str]
+    # Database settings  # Default to values from docker-compose.yml
+    DATABASE_HOST: Optional[str] = "db"
+    DATABASE_USER: Optional[str] = "jobbergate"
+    DATABASE_PSWD: Optional[str] = "local-pswd"
+    DATABASE_NAME: Optional[str] = "jobbergate"
     DATABASE_PORT: Optional[int] = 5432
-    DATABASE_URL: Optional[str]
+
```
matheushent, line 41: what's the benefit on building the URL by pieces? I guess it goes the other way around what we've been doing on the others APIs

matheushent: not that I will push hard against this way but I'd like some pattern being applied across our code

jobbergate-api/jobbergate_api/config.py
```diff
@@ -32,13 +32,19 @@ class Settings(BaseSettings):
...
+    # Test database settings
+    TEST_DATABASE_HOST: Optional[str] = "localhost"
+    TEST_DATABASE_USER: Optional[str] = "test"
+    TEST_DATABASE_PSWD: Optional[str] = "test-pswd"
+    TEST_DATABASE_NAME: Optional[str] = "test-jobbergate"
+    TEST_DATABASE_PORT: Optional[int] = 5433
```
matheushent, line 47: we could set this stuff in the docker-compose file and remove complexity from the code itself

jobbergate-api/jobbergate_api/main.py
```diff
@@ -4,24 +4,25 @@
...
+from jobbergate_api.config import DeployEnvEnum, settings
+from jobbergate_api.storage import database, handle_fk_error
 
 subapp = FastAPI()
 subapp.add_middleware(
     CORSMiddleware, allow_origins=["*"], allow_credentials=True, allow_methods=["*"], allow_headers=["*"],
 )
 
-if settings.SENTRY_DSN:
+if settings.SENTRY_DSN and settings.DEPLOY_ENV != DeployEnvEnum.TEST:
```
matheushent, line 25: `DeployEnvEnum.LOCAL` as well, doesn't it?

jobbergate-api/jobbergate_api/storage.py
```diff
@@ -1,24 +1,51 @@
...
+    return str(
+        URL.build(
+            scheme="postgresql",
+            user=getattr(settings, f"{prefix}DATABASE_USER"),
+            password=getattr(settings, f"{prefix}DATABASE_PSWD"),
+            host=getattr(settings, f"{prefix}DATABASE_HOST"),
+            port=getattr(settings, f"{prefix}DATABASE_PORT"),
+            path="/{}".format(getattr(settings, f"{prefix}DATABASE_NAME")),
+        )
+    )
```
matheushent, line 45: I can't see the benefit of this module against having simply one variable in the config module defining the database URL. IMO we should just move the test postgres vars to the `docker-compose.yaml`

matheushent (changes requested):

.github/workflows/test_on_push.yaml
```diff
@@ -17,6 +17,13 @@ jobs:
           POSTGRES_PASSWORD: test-pswd
           POSTGRES_USER: test
           POSTGRES_DB: test-jobbergate
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
```
jamesbeedy, line 24: cool!

jamesbeedy (dismissed):
👍

jobbergate-api/Makefile
```diff
@@ -34,17 +34,13 @@ format: install
...
-	poetry run uvicorn --host 0.0.0.0 jobbergate_api.main:app --reload
-
 .PHONY: publish
 publish:
 	../release-scripts/publish-release.sh
 
 .PHONY: local
 local: install
-	poetry run uvicorn --host 0.0.0.0 jobbergate_api.main:app --reload
+	poetry run jobbergate dev-server
```
jamesbeedy, line 43: I'm not sure we want to add dev tools to the `jobbergate` cli. Is there a better place to put the `dev-server` functionality? Or possibly there is a reason you put it here?

jamesbeedy, line 43: I'm not sure we want to add dev tools to the `jobbergate` cli. Is there a better place to put the `dev-server` functionality? Or possibly there is a reason you put it here?

Oh ... I think I see .... this is the `jobbergate` that is part of the jobbergate-api ....

In this case we have a naming clash with the `jobbergate` command exposed by the jobbergate-cli.

docker-compose.yml
dusktreader: They're just default values.  There isn't anything sensitive here. What would be the use of dummy values?

jobbergate-api/jobbergate_api/apps/applications/routers.py
```diff
@@ -43,21 +42,15 @@ async def applications_create(
...
-    async with database.transaction():
-        try:
-            insert_query = applications_table.insert()
-            values = application.dict()
-            inserted_id = await database.execute(query=insert_query, values=values)
-
-        except INTEGRITY_CHECK_EXCEPTIONS as e:
-            raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
+    try:
+        insert_query = applications_table.insert().returning(applications_table)
```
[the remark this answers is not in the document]
dusktreader: It does work. If you look at the columns returned when you use this expression, you'll see that it includes all the columns from the table.

docker-compose.yml
matheushent: I mean set fixed params like:

```shell
POSTGRES_PASSWORD=dummy-password
POSTGRES_USER=omnivector
POSTGRES_DB=jobbergate
```

It will avoid to set unnecessary env vars. You can ask me "ok but what's the problem on three more env vars?". The problem is the application itself is already complex so any new piece we add and increment the complexity makes it harder and harder to maintain. Does it make sense?

dusktreader: Having the defaults lets us override them with a `.env` file or your local environment.

jobbergate-api/jobbergate_api/cli/db.py
dusktreader: No, these commands in the cli are extremely useful. For example, you can run `db login` and log directly into the database with the settings you have configured for the app. This is very helpful when debugging data issues. Othewise, you have to:
* remember the command line to log into postgres
* fetch and copy all the values from your environment to do so.

For the `start` commands, these are nice if you want to quickly spin up a postgres database in docker for running unit tests and such.

These commands are converted from previous API projects I've worked on, and they proved very useful in development for myself and other devs on the team. I would like to keep them.

These commands are not used in production, they are only for helping developers. There is very little maintenance required here.

jobbergate-api/jobbergate_api/cli/dev_server.py
dusktreader: We have a docker-compose.yaml that also spins up db services if you prefer to use that approach.

jobbergate-api/jobbergate_api/config.py
dusktreader: Because it allows you to use the same environment variables to start the docker databases with docker-compose.yaml or the dev cli. For example, if you set your `DATABASE_USER`, `DATABASE_PSWD`, `DATABASE_NAME` in your environment, then `docker-compose.yaml` can spin the database up with those values as well as your local app using them.

jobbergate-api/jobbergate_api/config.py
dusktreader: These are set in the docker-compose. Having default values here doesn't add complexity in my opinion.

jobbergate-api/jobbergate_api/main.py
dusktreader: Not necessarily. It's good to test your sentry connection locally.

jobbergate-api/jobbergate_api/storage.py
dusktreader: This module also validates the URL.

Having separate settings for the test database makes it much safer. You have no chance of accidently connecting to a non-test database when running unit tests. If you don't separate them, there's a significant chance that you will accidently forget to set the database_url to connect to a test database.

jobbergate-api/Makefile
dusktreader: That's a good point. I'll rename the dev cli commands to make it clearer.

dusktreader: Renamed in latest commit.

docker-compose.yml
matheushent: that's exactly my point: there's no need to let them to be overwritten. It just adds complexity. I think we could just set whatever dummy value comes in mind and boom, ready to go

jobbergate-api/jobbergate_api/config.py
matheushent: what I mean is just to use the compose, nothing more. Otherwise, only complexity is added. It is quite complicated to be dropped on a project and to have to set a bunch of "no needed" vars

jobbergate-api/jobbergate_api/config.py
matheushent: I am lost to be honest. Some are set in the compose, some are not...

jobbergate-api/jobbergate_api/storage.py
matheushent: a test like [this one](https://github.com/omnivector-solutions/armada-api/blob/main/tests/conftest.py#L26-L33) could solve this risk.

matheushent: and that's where it converges with my previous comments: there are a bunch of variables to be set out there. A test like this one could make it simpler and then it was just needed to assign some dummy values as default for the database in the docker-compose.yaml file

dusktreader: I don't think we should hard-code the expected test URL.

The biggest reason to break up the database settings is so that the same environment variables can be used to get values when creating the databases (via compose, the helper scripts, docker, whatever) as are used to connect within the app. This makes it much easier to just set your environment, start up your database, and know that your API is going to connect to that database.

docker-compose.yml
jamesbeedy: @dusktreader can you describe a scenario when you set these environment variables to override the defaults?

jamesbeedy (changes requested):
Looks really nice.

Only two things ....

1) `jobbergate-api/jobbergate_api/cli/__init__.py`, `jobbergate-api/jobbergate_api/cli/db.py`, `jobbergate-api/jobbergate_api/cli/dev_server.py` are development tools. Is there a reason for maintaining these tools as part of `jobbergate-api/jobbergate_api/cli` proper?

2) Endpoint descriptions end up in the swagger docs. Please close the endpoint descriptions with periods so they read like a sentence when inspecting the docs.

docker-compose.yml
dusktreader: Here's a couple of scenarios:
* Running multiple dbs simultaneously so you need to map different ports (for example, running test dbs for license-manager-backend, jobbergate-api, armada-api)
* Running multiple instances of the same product's databases at the same time. For instance, If I wanted to have 2 versions of jobbergate (maybe I'm debugging an issue on the last released version while also working on new features that require a schema change)
* Connecting to a database hosted on another server (say connecting a local jobbergate with no database changes to the jobbergate database running in staging)

jobbergate-api/jobbergate_api/config.py
dusktreader: The idea is to have the defaults be usable immediately without having to use docker-compose or create a .env file. for example, with `dev-tools`, I can simply clone the repo and run these three commands:
* `poetry install`
* `poetry run dev-tools db start-test`
* `poetry run pytest`

docker-compose can be helpful and convenient, but I don't like having projects require it to do meaningful things.

jobbergate-api/jobbergate_api/config.py
dusktreader: The docker-compose.yaml needs some cleanup. Removing defaults here isn't a great approach, though, because it means you have to set environment variables before you can do anything with the project. By setting defaults on the database variables, I can spin up local and test databases with the dev-tools without having to set environment variables at all.

dusktreader:
> Looks really nice.
> 
> Only two things ....
> 
> 1. `jobbergate-api/jobbergate_api/cli/__init__.py`, `jobbergate-api/jobbergate_api/cli/db.py`, `jobbergate-api/jobbergate_api/cli/dev_server.py` are development tools. Is there a reason for maintaining these tools as part of `jobbergate-api/jobbergate_api/cli` proper?
> 2. Endpoint descriptions end up in the swagger docs. Please close the endpoint descriptions with periods so they read like a sentence when inspecting the docs.

1. No, these tools could be moved into a separate sub-package that isn't installed when the package is installed with pip
2. Will do.

docker-compose.yml
jamesbeedy: 👍

dusktreader:
> Looks really nice.
> 
> Only two things ....
> 
> 1. `jobbergate-api/jobbergate_api/cli/__init__.py`, `jobbergate-api/jobbergate_api/cli/db.py`, `jobbergate-api/jobbergate_api/cli/dev_server.py` are development tools. Is there a reason for maintaining these tools as part of `jobbergate-api/jobbergate_api/cli` proper?
> 2. Endpoint descriptions end up in the swagger docs. Please close the endpoint descriptions with periods so they read like a sentence when inspecting the docs.

Both points have been addressed in the latest commit.

jamesbeedy (approved):
LGTM 👍  Thanks!

PR #86, opened 2022-03-15
JOB-111 Migrated job submission to cluster-agent

dusktreader:
#### What
Migrated all job submission (to sbatch) logic to the cluster-agent.

#### Why
This will allow users to submit jobs for execution through the API only (no need to be logged into a cluster)

`Task`: https://app.clickup.com/t/18022949/JOB-111

---

#### Peer Review


jobbergate-api/jobbergate_api/apps/job_submissions/constants.py
```diff
@@ -0,0 +1,7 @@
+from auto_name_enum import AutoNameEnum, NoMangleMixin, auto
+
+
+class JobSubmissionStatus(AutoNameEnum, NoMangleMixin):
+    CREATED = auto()
+    SUBMITTED = auto()
+    COMPLETED = auto()
```
jamesbeedy, line 7: Can you help me understand what this extra components are used for?

Could this simple be

```python
class JobSubmissionStatus(AutoNameEnum, NoMangleMixin):
    CREATED = "CREATED"
    SUBMITTED = "SUBMITTED"
    COMPLETED = "COMPLETED"
```
?

jamesbeedy (changes requested):
Only one objection concerning the `auto-name-enum` imports.

Resolve my comment and I will be +1 on this.

Thanks!

jamesbeedy:
One other comment concerning the new endpoints.....

For example, 
`"/job-submissions/agent/pending"`

Is it possible that the agent send a request to `/job-submissions` with arguments that filter for the specific status instead of making separate endpoints specifically for the agent that return job submissions for each status?

dusktreader:
> One other comment concerning the new endpoints.....
> 
> For example, `"/job-submissions/agent/pending"`
> 
> Is it possible that the agent send a request to `/job-submissions` with arguments that filter for the specific status instead of making separate endpoints for the agent that return job submissions for each status?

We discussed this in a conversation a day or two ago. I noted that the agent endpoints combine data via a join so that the agent need only make a single request to get job-submission, job-script, and application info in one request.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -186,6 +207,85 @@ async def job_script_update(job_submission_id: int, job_submission: JobSubmissio
...
+    token_payload: TokenPayload = Depends(guard.lockdown(Permissions.JOB_SUBMISSIONS_VIEW)),
+):
+    """
+    Get a list of pending job submissions for the cluster-agent.
+    """
+    if token_payload.client_id is None:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Access token does not contain a `client_id`. Cannot fetch pending submissions",
+        )
```
matheushent, line 226: This parameter is mandatory so I guess we could put it as required in the input model and let FastAPI handles

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -186,6 +207,85 @@ async def job_script_update(job_submission_id: int, job_submission: JobSubmissio
...
+                job_submissions_table.c.job_submission_name,
+                job_scripts_table.c.job_script_name,
+                job_scripts_table.c.job_script_data_as_string,
+                applications_table.c.application_name,
+            ]
+        )
+        .select_from(job_submissions_table.join(job_scripts_table).join(applications_table))
+        .where(job_submissions_table.c.status == JobSubmissionStatus.CREATED,)
+        .where(job_submissions_table.c.cluster_client_id == token_payload.client_id,)
+    )
```
matheushent, line 241: this is interesting implementation. I wonder if we could improve it with GraphQL.. anyway, just thinking out loud here

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -186,6 +207,85 @@ async def job_script_update(job_submission_id: int, job_submission: JobSubmissio
...
+    Update a job_submission with a new status.
+
+    Make a put request to this endpoint with the new status to update a job_submission.
+    """
+    client_id = token_payload.client_id
+    if client_id is None:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Access token does not contain a `client_id`. Cannot update job_submission",
+        )
```
matheushent, line 269: same stuff here

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -186,6 +207,85 @@ async def job_script_update(job_submission_id: int, job_submission: JobSubmissio
...
+    rows = await database.fetch_all(query)
+    return rows
+
+
+@router.put(
+    "/job-submissions/agent/{job_submission_id}",
+    status_code=200,
+    description="Endpoint for an agent to update the status of a job_submission",
+    response_model=JobSubmissionResponse,
+)
```
matheushent, line 252: I think this route should be a PATCH method and raise an error in case the job submissions ID doesn't match any real job submissions

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
```diff
@@ -44,6 +53,7 @@ class JobSubmissionCreateRequest(BaseModel):
     job_submission_description: Optional[str]
     job_script_id: int
     slurm_job_id: Optional[int]
+    cluster_client_id: Optional[str]
```
matheushent, line 56: according to my previous comments: make it mandatory is needed

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
```diff
@@ -44,6 +53,7 @@ class JobSubmissionCreateRequest(BaseModel):
     job_submission_description: Optional[str]
     job_script_id: int
     slurm_job_id: Optional[int]
```
matheushent, line 55: also, this is not needed in the moment of creation because the agent updates it later so this parameter should be passed only when updating the job submissions

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
```diff
@@ -56,6 +66,7 @@ class JobSubmissionUpdateRequest(BaseModel):
 
     job_submission_name: Optional[str]
     job_submission_description: Optional[str]
+    status: Optional[JobSubmissionStatus]
```
matheushent, line 69: `slurm_job_id` is needed, doesn't it?

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -186,6 +207,85 @@ async def job_script_update(job_submission_id: int, job_submission: JobSubmissio
...
+    query = (
+        select(
+            columns=[
+                job_submissions_table.c.id,
+                job_submissions_table.c.job_submission_name,
+                job_scripts_table.c.job_script_name,
+                job_scripts_table.c.job_script_data_as_string,
+                applications_table.c.application_name,
+            ]
+        )
```
matheushent, line 237: I'm wondering if we could since select the rows that don't have the `slurm_job_id` set because this parameter should be set only when the agent updates it. It seems a more efficient implementation for me

matheushent (changes requested):

jobbergate-api/jobbergate_api/apps/job_submissions/models.py
```diff
@@ -16,6 +17,8 @@
     Column("job_submission_owner_email", String, nullable=False, index=True),
     Column("job_script_id", ForeignKey("job_scripts.id"), nullable=False),
     Column("slurm_job_id", Integer, default=None),
+    Column("cluster_client_id", String, nullable=False, index=True),
+    Column("status", Enum(JobSubmissionStatus), nullable=False, index=True),
```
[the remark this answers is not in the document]
dusktreader: Because we filter on status in some queries. This will boost the performance as the data scales.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: Unless we changed the data backend, we would still end up with something fundamentally like this. I think it might be worth exploring changing the api layer and the data layer at some point in the future.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: There's no parameter on this endpoint...

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: If you allow an enum column to be nullable, you essentially add an additional implict value to the enum (null). Then you have an issue where trying to understand the status of a specific row means you have to understand the implementation fully (what does null mean?).

I prefer to make enum columns non-nullable and for the rows to have an explicit status.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: See my comment above.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: If the job_submission_id doesn't match an existing row, the update database query will nave a null result which causes a 404 being raised.

It does make sense for it to be a PATCH instead of PUT, however.

jobbergate-api/job_script_example.py
```diff
@@ -0,0 +1,477 @@
...
+            print(f'Abaqus exited with ERROR CODE {abaqus_result}', flush=True)
+            exit_status = "Exited"
+    except NameError:
+        print("Could not fetch Abaqus' error code.")
+        exit_status = "Exited"
+    # Stage back result files files from local scratch
+    _stageback()
+finally:
+    _logging()
+    _cleanup()"
```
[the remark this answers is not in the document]
matheushent: I suggest to put in an example folder. It could be helpful for incoming maintainers

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -41,17 +45,28 @@ async def job_submission_create(
...
+
     create_dict = dict(
-        **job_submission.dict(exclude_unset=True), job_submission_owner_email=identity_claims.user_email,
+        **job_submission.dict(exclude_unset=True),
+        job_submission_owner_email=identity_claims.user_email,
+        status=JobSubmissionStatus.CREATED,
     )
+    if job_submission.cluster_client_id is None:
+        create_dict.update(cluster_client_id=token_payload.client_id)
```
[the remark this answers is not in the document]
matheushent: I shall require to make the incoming cluster id only from the input body. Having two differents ways of supplying the same thing can lead to unexpected behaviours

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
dusktreader: Good point!

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
matheushent: client ID I mean. IMO, the best approach is to make it mandatory

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
matheushent: I see. So what about just changing this one and change the others in the future?

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
dusktreader: Addressed in latest commit.

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
dusktreader: We will make adjustments to this mechanism in a future task.

jobbergate-api/job_script_example.py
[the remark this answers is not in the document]
dusktreader: Created an example root folder and put this and the dummy application there as well.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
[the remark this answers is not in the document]
dusktreader: We will make adjustments to this mechanism in a future task.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: We will investigate this more in a future task.

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
dusktreader: We will make adjustments to this mechanism in a future task.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: I will create a task to do this in a future sprint.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
dusktreader: We will make adjustments to this mechanism in a future task.

matheushent (approved):

PR #98, opened 2022-04-12
Added some examples of using the API, tweaked the API a bit, updated …

dusktreader:
#### What
Added some examples of using the API directly to get auth tokens and create job-scripts and job-submissions.
Also updated the API so that it can use the uploaded application's config as the default for param_dict. Thus, a job-script created through the API may default to values already provided in the application.

---

#### Peer Review


jamesbeedy (approved):
👍

PR #104, opened 2022-05-04
ARMADA-517 Added graceful handling of pyperclip exceptions

dusktreader:
#### What
Absorbs exceptions from pyperclip gracefully.

#### Why
The ability to copy tokens to the clipboard is very nice, but it breaks when the system does not have a clipboard set up. This absorbs the exception without removing pyperclip.

`Task`: https://app.clickup.com/t/18022949/ARMADA-517

---

#### Peer Review

### DIFF
--- a/jobbergate-cli/jobbergate_cli/main.py
+++ b/jobbergate-cli/jobbergate_cli/main.py
@@ -7,7 +7,6 @@ from typing import Optional
 import httpx
 import importlib_metadata
 import jose
-import pyperclip
 import typer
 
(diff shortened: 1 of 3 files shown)


matheushent (commented):
Instead, why simply not try-except when there's no clipboard available and print it anyway?

dusktreader:
> Instead, why simply not try-except when there's no clipboard available and print it anyway?

Couple of reasons for this implementation. First, putting it in a function makes it testable, which is nice. Second, we want to only show the footer text "token copied to clipboard" _if_ pyperclip was able to do that. If it fails, we still print it but don't show that footer.

matheushent (approved):

PR #108, opened 2022-05-25
Making file lister case-insensitive

NilesDev:
#### What
Makes the helper function get_file_list ignore case when searching for files. 
Could (maybe should?) be made optional to not change function of existing applications. Only difference should be the flag in re.compile: 0 vs re.IGNORECASE.

#### Why
More flexible searching.

---

#### Peer Review


dusktreader:
This looks like a good change. I'll fix up the unit tests and also back-port this to the legacy jobbergate as well next week.

fschuch (approved):

PR #110, opened 2022-06-06
Set `environment` variable when Sentry initializes

matheushent:
#### What
This PR only adds a parameter to the settings class and use it as Sentry's environment variable

#### Why
Needed to better organize Sentry

`Task`: https://app.clickup.com/t/18022949/ARMADA-528

---

#### Peer Review

### DIFF
--- a/jobbergate-cli/jobbergate_cli/config.py
+++ b/jobbergate-cli/jobbergate_cli/config.py
@@ -26,6 +26,7 @@ class Settings(BaseSettings):
     # enable http tracing
     JOBBERGATE_DEBUG: bool = Field(False)
 
+    # Setry's configuration
     SENTRY_DSN: Optional[str]
     SENTRY_TRACE_RATE: float = Field(1.0, gt=0.0, le=1.0)
     SENTRY_ENV: str = "local"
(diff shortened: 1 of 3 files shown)


jobbergate-cli/jobbergate_cli/config.py
```diff
@@ -28,6 +28,7 @@ class Settings(BaseSettings):
 
     SENTRY_DSN: Optional[str]
     SENTRY_TRACE_RATE: float = Field(1.0, gt=0.0, le=1.0)
+    SENTRY_ENV: str = "local"
```
fschuch, line 31: A comment in the beginning of the block would be good to maintain consistency:

```suggestion
    # Sentry <resume functionality here>
    SENTRY_DSN: Optional[str]
    SENTRY_TRACE_RATE: float = Field(1.0, gt=0.0, le=1.0)
    SENTRY_ENV: str = "local"
```

I'm just not sure about the exacly sentence, what do you think?

fschuch (changes requested):
Good job, just one comment to be addressed

jobbergate-cli/jobbergate_cli/config.py
matheushent: sure, sounds good

fschuch (approved):

PR #111, opened 2022-06-08
Armada-460 move job script files into s3

fschuch:
#### What
Drop the column `job_script_data_as_string` from `job_scripts` table. They should be put in S3 just like the application files.

#### Why
We are currently storing the files in our database. There have been issues with this around moving the files back and forth between the CLI and the API.

`Task`: https://app.clickup.com/t/18022949/ARMADA-460

---

#### Peer Review


jobbergate-api/jobbergate_api/tests/apps/job_scripts/conftest.py
```diff
@@ -59,19 +57,3 @@ def template_files(template_files_application_content):
     Provide a fixture that creates test template files.
     """
     return {"application.sh": template_files_application_content}
-
-
-@fixture
-def s3_object():
```
fschuch, line 65: Check if we really using this fixture in all places where we should.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
```diff
@@ -270,6 +269,8 @@ async def job_script_delete(job_script_id: int = Query(..., description="id of t
         )
 
     delete_query = job_scripts_table.delete().where(where_stmt)
+    s3man_jobscripts.delete(job_script_id)
```
fschuch, line 272: put it in a try/except block and catch `KeyError`, in case job script was not found (i.e., already deleted)

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
```diff
@@ -287,21 +288,32 @@ async def job_script_update(job_script_id: int, job_script: JobScriptUpdateReque
...
+        )
         .returning(job_scripts_table)
     )
     try:
         result = await database.fetch_one(update_query)
     except INTEGRITY_CHECK_EXCEPTIONS as e:
         raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
+    else:
+        if job_script.job_script_data_as_string:
+            s3man_jobscripts.put(
```
fschuch, line 302: Should we update the database if any error occurs when uploading the job script to S3?

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -40,32 +160,60 @@ def _get_key(self, app_id):
...
-        self.s3_client.put_object(
-            Body=upload_file.file,
-            Bucket=self.bucket_name,
-            Key=key,
-        )
+        logger.debug(f"Uploading to S3: {key})")
+        if isinstance(upload_file, str):
+            self.client.put(upload_file, key)
+        else:
+            self.client.put(upload_file.file, key)
```
fschuch, line 172: This block may be unnecessary if we use the S3 fixture properly. Investigate it.

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -40,32 +160,60 @@ def _get_key(self, app_id):
...
         """
-        return self.s3_client.get_object(Bucket=self.bucket_name, Key=self._get_key(app_id))
+        key = self._get_key(app_id)
+        logger.debug(f"Getting from S3: {key})")
+        return self.client.get(key=key)
+
+    def get_s3_object_as_string(self, app_id) -> str:
+        s3_file_obj = self.get(app_id)
+        if isinstance(s3_file_obj, str):
+            return s3_file_obj
```
fschuch, line 194: This block may be unnecessary if we use the S3 fixture properly. Investigate it.

jobbergate-api/jobbergate_api/apps/job_scripts/schemas.py
```diff
@@ -89,20 +89,33 @@ class Config:
...
+    """
+    Complete model to match database in addition to the field
+    job_script_data_as_string from S3, for the JobScript resource.
+    """
+
+    job_script_data_as_string: str
+
+    class Config:
+        orm_mode = True
+        schema_extra = job_script_meta_mapper
```
fschuch, line 121: Remove `job_script_data_as_string` from some endpoints to reduce the response size, in line with what we did for applications. It is just included when getting information about specific job script id.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
fschuch: @matheushent @dusktreader

jobbergate-api/jobbergate_api/apps/job_scripts/schemas.py
fschuch: @matheushent @dusktreader

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
```diff
@@ -191,8 +175,19 @@ async def job_script_create(
     except INTEGRITY_CHECK_EXCEPTIONS as e:
         raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=str(e))
 
+    else:
+
+        s3man_jobscripts.put(
```
dusktreader, line 180: I have a couple of issues with this:
First, there's no exception handling around the `put()` method here. What happens if the user doesn't have write permission on the bucket or any other s3 access error? I think we will need a column to indicate if the files were successfully uploaded to s3 or not. Otherwise, we could easily have what appears to be a valid job_script but that has no script uploaded.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
dusktreader: Since the script itself is the core of the entries  in "job_scripts", I think it makes sense to actually roll-back the database changes if the script could not be uploaded. I can't think of a case where it's very useful to have an entry in job_scripts but the script itself is not there.

I think we should do the same for the `create` logic as well. In a try/except block, create the row and upload the data. If any of the operations fails, rollback the database and throw an exception that provides the context of what went wrong.

jobbergate-api/jobbergate_api/apps/job_scripts/schemas.py
dusktreader: This approach makes sense for now. We should think more about whether it makes sense to have a separate endpoint for pulling down the script itself (as opposed to its metadata).

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -1,37 +1,157 @@
...
+    @abstractmethod
+    def get(self, key: str) -> typing.IO:
+        """
+        Get a file from the client associated to the given key.
+        Raise KeyError if the file does not exist.
+        """
+        pass
+
+
+class DummyClient(BucketClientBase):
```
dusktreader, line 46: I'm not sure why this is necessary. Why not just mock the s3 client in the tests?

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -1,37 +1,157 @@
...
+
+        if client is None:
+            client = S3Client()
+
+        self.client = client
+
+    def _get_key(self, app_id: str) -> str:
         """
-        Get an s3 key based upon the app_id. If app_id is falsey, throw an exception.
+        Get an s3 key based upon the app_id. If app_id is empty, throw an exception.
```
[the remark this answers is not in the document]
dusktreader: Well, in this case....perhaps it's better as `empty`. Since the type-hint suggests that `app_id` should only be a string and because only an empty string resolves as `False`, this is more narrowly correct than saying `falsey`.

jobbergate-api/jobbergate_api/s3_manager.py
dusktreader: 💯 

The type-hint says `upload_file` is an instance of `UploadFile`. Are we actually sending this method a string anywhere in our code-base?

dusktreader (changes requested):
Made some comments for you. Nice work thus far!

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
matheushent: well, this log wouldn't be 100% precise because it doesn't mean the the job is already deleted; the job could never exist in our database. Line 271 won't cause error in case we try to delete a record that doesn't exist, so, I think we just put in a try-except block and return 404 in case we see `KeyError`

jobbergate-api/jobbergate_api/tests/apps/job_scripts/test_routers.py
```diff
@@ -143,7 +142,7 @@ async def test_create_job_script(
     assert job_script.job_script_owner_email == "owner1@org.com"
     assert job_script.job_script_description is None
     assert job_script.job_script_data_as_string
-    assert job_script.job_script_data_as_string != job_script_data["job_script_data_as_string"]
+    assert job_script.job_script_data_as_string.startswith('{"application.sh":')
```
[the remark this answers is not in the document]
fschuch: Yeah, very good point, no need for it.

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -40,32 +160,60 @@ def _get_key(self, app_id):
...
+        s3_application_tar = tarfile.open(fileobj=BytesIO(s3_application_obj["Body"].read()))
+        return s3_application_tar
+
+
+s3_client = S3Client()
+s3man_applications = S3Manager(client=s3_client)
+s3man_jobscripts = S3Manager(
+    directory_name="job-scripts",
+    filename="job-script.json",
+    client=s3_client,
```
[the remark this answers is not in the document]
fschuch: I agree, fix included in next commit

jobbergate-api/jobbergate_api/s3_manager.py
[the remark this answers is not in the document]
fschuch: Ok, no need to change.

jobbergate-api/jobbergate_api/s3_manager.py
fschuch: Yes, some tests send strings. I'm working on it.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
fschuch: Absolutely.
Mental note: Write tests to verify the roll-back.

jobbergate-api/jobbergate_api/s3_manager.py
fschuch: For me, `DummyClient` makes testing and debugging easier, and test files are also a bit cleaner without the mocks. I'll keep this suggestion in mind while solving the other ones.

jobbergate-api/jobbergate_api/apps/job_scripts/schemas.py
fschuch: Yeah, let's keep the idea of a new endpoint to recover the script specifically.

Solving the conversation, no changes are needed for now.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
fschuch: Yeah, I'll implement a try/except and throw an HTTP exception if `s3man_jobscripts.put` goes wrong.

I think that a status column is not necessary if we implement a roll-back in the database when the endpoints create and put face errors with S3, as discussed in other points in this review.

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -1,37 +1,157 @@
...
+    def __init__(
+        self,
+        directory_name: str = "applications",
+        filename="jobbergate.tar.gz",
+        client: BucketClientBase = None,
+    ):
+        """
+        Initialize the S3Manager class instance.
+        """
+        self.key_template = directory_name + "/{app_id}/" + filename
```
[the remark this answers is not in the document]
fschuch: Good point. Fixed in the next commit.

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -1,37 +1,157 @@
...
+            Key=key,
+        )
+
+    def delete(self, key: str) -> None:
+        """
+        Delete a file from the client associated to the given key.
+        """
+        try:
+            self.s3_client.delete_object(Bucket=self.bucket_name, Key=key)
+        except self.s3_client.exceptions.NoSuchBucket:
```
fschuch, line 115: `NoSuchBucket` could be tested at initialization. Should we create a new bucket in that case?

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
matheushent: I think this operation in the time consuming one. That said, it should be done first than the SQL query, this way we wouldn't spend unnecessary SQL queries in case there's error uploading to S3

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
```diff
@@ -287,21 +288,32 @@ async def job_script_update(job_script_id: int, job_script: JobScriptUpdateReque
...
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
             detail=f"JobScript with id={job_script_id} not found.",
         )
 
-    return result
+    job_script_response = dict(result)
+    job_script_response["job_script_data_as_string"] = s3man_jobscripts.get_s3_object_as_string(job_script_id)
+
+    return job_script_response
```
matheushent, line 316: Theorically it a bit strange for me the route method is PUT. If we throw 404 then it should be PATCH, shouldn't?

matheushent:
Question: will the job scripts be in the same bucket as the applications?

jobbergate-api/jobbergate_api/s3_manager.py
fschuch: > Question: will the job scripts be in the same bucket as the applications?

They are for now. But we can change it if necessary. Do you think separated buckets are better?

jobbergate-api/jobbergate_api/s3_manager.py
fschuch: Fair point, but I've seen this error sometimes when testing locally. Is that a way to create the bucket from docker-compose?

fschuch:
> Question: will the job scripts be in the same bucket as the applications?

They are for now. But we can change it if necessary. Do you think separated buckets are better?

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
fschuch: The is a query and a test a few lines above (GitHub is not showing here) that raises 404 if the id is not in the database, so the log suggested should be enought.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
fschuch: Well, I also see the opposite. If we upload first to S3 and something goes wrong with the database, we will have problems rolling back the file at S3 (it is not possible, as far as I know).

jobbergate-api/jobbergate_api/s3_manager.py
fschuch: solved in [75b4341](https://github.com/omnivector-solutions/jobbergate/pull/111/commits/75b434114eb0bbfac0917da9c87101bb9c9eae79)

jobbergate-api/jobbergate_api/tests/test_s3_manager.py
```diff
@@ -0,0 +1,55 @@
...
+        ("applications", "applications/{app_id}/jobbergate.tar.gz"),
+        ("job-script", "job-script/{app_id}/jobbergate.tar.gz"),
+    ],
+)
+def test_s3_manager__key_template(directory_name, desired_template):
+    s3man = S3Manager(directory_name=directory_name, client=DummyClient())
+    assert s3man.key_template == desired_template
+
+
+s3man = S3Manager()
```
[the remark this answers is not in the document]
fschuch: solved in [75b4341](https://github.com/omnivector-solutions/jobbergate/pull/111/commits/75b434114eb0bbfac0917da9c87101bb9c9eae79)

jobbergate-api/jobbergate_api/tests/test_s3_manager.py
```diff
@@ -0,0 +1,55 @@
+from unittest import mock
+
+import pytest
+from botocore.exceptions import BotoCoreError
+from fastapi.exceptions import HTTPException
+
+from jobbergate_api.s3_manager import DummyClient, S3Manager
```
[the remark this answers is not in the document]
fschuch: The were removed in solved in [75b4341](https://github.com/omnivector-solutions/jobbergate/pull/111/commits/75b434114eb0bbfac0917da9c87101bb9c9eae79), so no need for tests

jobbergate-api/jobbergate_api/s3_manager.py
fschuch: Good points around `DummyClient`, it was removed in [75b4341](https://github.com/omnivector-solutions/jobbergate/pull/111/commits/75b434114eb0bbfac0917da9c87101bb9c9eae79)

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -40,32 +160,60 @@ def _get_key(self, app_id):
...
+        return self.client.get(key=key)
+
+    def get_s3_object_as_string(self, app_id) -> str:
+        s3_file_obj = self.get(app_id)
+        if isinstance(s3_file_obj, str):
+            return s3_file_obj
+        string = s3_file_obj.get("Body").read().decode("utf-8")
+        return string
+
+    def get_s3_object_as_tarfile(self, app_id):
```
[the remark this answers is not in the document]
fschuch: After some investigation, it looks like the files are stored in S3 as text files. The API converts them to tarfiles internally before sending the information to:
https://github.com/omnivector-solutions/jobbergate/blob/c8c9f210e447c47626c0105660e814595f2fdb30/jobbergate-api/jobbergate_api/apps/job_scripts/routers.py#L85

I"ll keep it as it is for now since it is not the goal of this PR.

jobbergate-api/jobbergate_api/tests/apps/job_scripts/test_routers.py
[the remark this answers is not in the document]
fschuch: Would `assert isinstance(job_script.job_script_data_as_string, str)` be better? Because the original comparison is not possible anymore, `job_script_data` does not contain the key `"job_script_data_as_string"`.

jobbergate-api/jobbergate_api/s3_manager.py
fschuch: Exceptions for `NoSuchBucket` were removed in [75b4341](https://github.com/omnivector-solutions/jobbergate/pull/111/commits/75b434114eb0bbfac0917da9c87101bb9c9eae79)

dusktreader:
Task linked: [ARMADA-460 Move job_script files into S3](https://app.clickup.com/t/18022949/ARMADA-460)

jobbergate-api/jobbergate_api/tests/apps/job_scripts/test_routers.py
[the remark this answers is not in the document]
dusktreader: Meh, I think it's not that important. If it's truthy, that's probably good enough for this test.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
fschuch: The logic to roll-back the database was implemented.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
fschuch: This operation is now done in a try-except block, providing a message if anything goes wrong. The exception handling for no such bucket was removed.

jobbergate-api/jobbergate_api/tests/apps/job_scripts/test_routers.py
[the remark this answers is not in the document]
fschuch: This assertion was improved.

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -40,32 +105,51 @@ def _get_key(self, app_id):
...
+        in the expected format.
         """
-        key = self._get_key(app_id)
         try:
-            self.s3_client.delete_object(Bucket=self.bucket_name, Key=key)
-        except self.s3_client.exceptions.NoSuchBucket:
-            raise KeyError(f"No such bucket: {self.bucket_name}")
-        except self.s3_client.exceptions.NoSuchKey:
-            raise KeyError(f"No such key: {key}")
+            return self.s3_client.list_objects(
```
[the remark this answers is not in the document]
fschuch: Problem fixed.

jobbergate-api/jobbergate_api/tests/apps/job_scripts/conftest.py
```diff
@@ -59,19 +57,3 @@ def template_files(template_files_application_content):
...
-    """
-    Provide a fixture that creates a test s3 object.
-    """
-    return {"Body": open("jobbergate_api/tests/apps/job_scripts/test_files/jobbergate.tar.gz", "rb")}
```
fschuch, line 69: These fixtures were deleted because they were not used anywhere. In this case, we should consider deleting the file mentioned here as well.

fschuch: Actually, the fixtures were moved to another file.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
```diff
@@ -181,18 +165,32 @@ async def job_script_create(
...
 
-    except INTEGRITY_CHECK_EXCEPTIONS as e:
-        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=str(e))
+    async with database.transaction():
+
+        try:
+            insert_query = job_scripts_table.insert().returning(job_scripts_table)
+            job_script_data = await database.fetch_one(query=insert_query, values=create_dict)
+
+            assert job_script_data is not None
```
matheushent, line 178: shouldn't 400 be returned instead of AssertionError? This AssertionError could be hard to debug

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
```diff
@@ -287,21 +302,40 @@ async def job_script_update(job_script_id: int, job_script: JobScriptUpdateReque
...
+        try:
+            result = await database.fetch_one(update_query)
+        except INTEGRITY_CHECK_EXCEPTIONS as e:
+            raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
+
+        if result is None:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail=f"JobScript with id={job_script_id} not found.",
+            )
```
matheushent, line 322: in terms of the HTTP protocol, as far as I understand, a PUT method should never return 404. In this case the method bust be PATCH. Do you agree? @dusktreader might have some input as well

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -1,69 +1,214 @@
...
 
-    def _get_key(self, app_id):
+    def __iter__(self) -> typing.Iterator[str]:
         """
-        Get an s3 key based upon the app_id. If app_id is falsey, throw an exception.
+        Yield all ids found in the work folder.
         """
-        if not app_id:
+        for key in self._get_list_of_objects():
+            yield self._get_app_id_from_key(key)
```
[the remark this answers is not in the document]
matheushent: ahhhh nevermind, I saw you used it, sorry.

jobbergate-api/jobbergate_api/s3_manager.py
```diff
@@ -1,69 +1,214 @@
...
+    s3_application_tar = tarfile.open(fileobj=BytesIO(s3_application_obj["Body"].read()))
+    return s3_application_tar
+
+
+s3_client = boto3.client(
+    "s3",
+    endpoint_url=settings.S3_ENDPOINT_URL,
+)
+s3man_applications = S3ManagerRaw(s3_client, "applications", "jobbergate.tar.gz")
+s3man_jobscripts = S3ManagerText(s3_client, "job-scripts", "jobbergate.txt")
```
matheushent, line 214: this module is beautiful work mate, this is insane

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
matheushent: mypy...... such a boring thing sometimes

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
```diff
@@ -272,6 +281,12 @@ async def job_script_delete(job_script_id: int = Query(..., description="id of t
     delete_query = job_scripts_table.delete().where(where_stmt)
     await database.execute(delete_query)
 
+    try:
+        del s3man_jobscripts[job_script_id]
+    except KeyError:
+        # There is no need to raise an error if we try to delete a file that does not exist
+        logger.warning(f"Tried to delete job_script={job_script_id}, but it was not found")
+
```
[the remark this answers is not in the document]
dusktreader: No, because the delete request is issued for a row in the database. If that is successfully deleted, then the keyerror is incidental if the file was removed on s3 already.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
dusktreader: A PUT can definitely return a 404. If you give a valid body but a bad key (the id for the entity to be updated), this should absolutely be a 404.

A 404 should never be returned for a POST.

The only difference between a PUT and a PATCH is that a PUT expects to update an entire existing entity while a PATCH may only partially update an existing entity.

matheushent: I'm not sure if I'm on this page. If there's no key corresponding when the method is PUT then the router would create the key. PATCH would return 404

dusktreader: We discussed this. Resolving for now, conversation can carry on separately.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
dusktreader: We discussed this separately. Maybe we address this in the future (conceptual difference between PUT and PATCH). Let's just move forward as-is for now.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
fschuch: Comment adressed at [5624142](https://github.com/omnivector-solutions/jobbergate/pull/111/commits/56241424c53fcd0e8abcbb7fd51ec1508c84ff8d).

dusktreader (approved):

PR #115, opened 2022-06-24
Adjusted for migration to keycloak

dusktreader:
#### What
Adjusted jobbergate for migration to keycloak


jobbergate-cli/jobbergate_cli/subapps/clusters/tools.py
```diff
@@ -0,0 +1,111 @@
...
+            support=True,
+            json=dict(
+                query="query {cluster{clientId}}",
+                variables=dict(),
+            ),
+        ),
+    )
+
+    try:
+        cluster_names = [e["clientId"] for e in response_data["data"]["cluster"]]
```
matheushent, line 39: there are difference between the client ID and the name. They both exist in the Cluster API and they are different. Both are unique. So, you can search in the GraphQL query by the `clientId` and ask for the `name` field

jobbergate-cli/jobbergate_cli/subapps/clusters/tools.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def get_cluster_names(ctx: JobbergateContext) -> List[str]:
+    assert ctx.client is not None
+
+    cluster_names = load_clusters_from_cache()
+    if cluster_names is None:
+        cluster_names = pull_cluster_names_from_api(ctx)
+        save_clusters_to_cache(cluster_names)
+    print("CLUSTER NAMES: ", cluster_names)
```
matheushent, line 90: leftover? I guess just changing `print` for `logger.debug` would be good as well

jobbergate-cli/jobbergate_cli/subapps/clusters/tools.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if datetime.utcnow().timestamp() - cache_data.updated_at.timestamp() > settings.JOBBERGATE_CLUSTER_CACHE_LIFETIME:
+        logger.warning("Cached cluster data is expired")
+        return None
+
+    return cache_data.cluster_names
+
+
+def get_cluster_names(ctx: JobbergateContext) -> List[str]:
+    assert ctx.client is not None
```
matheushent, line 84: what's the use case for this to be `None`?

matheushent (changes requested):
Two points left:

1. It seems there are some missunderstanding between a cluster name and client Id. They are different and are mapped differently in the Cluster API;
2. Missing CHANGELOG and package version update.

Beyond that, good job mapping these changes 👍🏻

jobbergate-cli/jobbergate_cli/subapps/clusters/tools.py
dusktreader: This is an assertion to satisfy mypy/pylance.

jobbergate-cli/jobbergate_cli/subapps/clusters/tools.py
dusktreader: Yeah, leftovers. Removed in commit 5af24fc

jobbergate-cli/jobbergate_cli/subapps/clusters/tools.py
dusktreader: Adjusted in commit 5af24fc

matheushent (approved):

PR #132, opened 2022-08-12
Add back the action to publish to ecr

fschuch:
#### What
Add back the action to publish to ecr.

#### Why
We use it to deploy the api.

`Task`: https://omnivector-solutions.monday.com/boards/XXXX/pulses/YYYY

---

#### Peer Review

### DIFF
--- a/.github/workflows/publish-to-ecr.yaml
+++ b/.github/workflows/publish-to-ecr.yaml
@@ -1,0 +1,70 @@
+name: Publish image to ECR
+
+on:
+  workflow_dispatch:
+    inputs:
+      repository:
+        description: Repository environment name
+        default: 'jobbergate-api-staging'
+        required: true
+      region:
+        description: AWS region to which upload the package
+        required: true
+        default: 'us-west-2'
+
+jobs:
+  publish:
+    name: Publish to ECR
+    runs-on: ubuntu-latest
+    steps:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-ecr.yaml
```diff
@@ -0,0 +1,70 @@
...
+        run: |
+          echo "Poetry version is: $(poetry version --short)"
+          echo "::set-output name=version::$(poetry version --short)"
+
+      - name: Fail if poetry package version doesn't match tag
+        if: ${{ github.ref_name != steps.poetry-package-version.outputs.version }}
+        run: |
+          echo "Poetry package version doesn't match tag!"
+          echo "tag=${{ github.ref_name }}, version=${{ steps.poetry-package-version.outputs.version }}"
+          exit 1
```
matheushent, line 48: do you mind removing this? It makes harder to publish alpha and beta versions since it'd require unsignificant commits

matheushent: but if you prefer to keep the way it is, np as well

matheushent (approved):

.github/workflows/publish-to-ecr.yaml
fschuch: We agreed to keep this test after further discussion.

PR #139, opened 2022-09-09
Patch application files

fschuch:
#### What

* Template files were not included in the response of get_application_by_id;
* One call at the email notification system was not inside the error handler, resulting in an error when from_email was not defined on the configuration file;
* The CLI relies on the `application_list` endpoint to select one by identifier, but this endpoint does not return the application files anymore. It should call get_application_by_id  instead.
* Jobbergate was not receiving the report_message from cluster-agent when a job submission was rejected;
* A few improvements on log messages.

#### Why

A detailed round of tests and fixes for the new release.

---

#### Peer Review

### DIFF
--- a/jobbergate-api/jobbergate_api/apps/applications/routers.py
+++ b/jobbergate-api/jobbergate_api/apps/applications/routers.py
@@ -1,7 +1,7 @@
 """
 Router for the Application resource.
 """
-from typing import List, Optional
+from typing import List, Optional, Union
 
 from armasec import TokenPayload
 from fastapi import APIRouter, Depends, File, Header, HTTPException, Query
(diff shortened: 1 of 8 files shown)


jobbergate-api/jobbergate_api/apps/applications/routers.py
```diff
@@ -260,23 +260,36 @@ async def applications_list(
...
+    Return the application given it's id (when int) or identifier (when str).
     """
-    logger.debug(f"Getting {application_id=}")
+    logger.debug(f"Getting {application_identification=}")
+
+    if isinstance(application_identification, int):
+        query = applications_table.select().where(applications_table.c.id == application_identification)
+    elif isinstance(application_identification, str):
+        query = applications_table.select().where(
+            applications_table.c.application_identifier == application_identification
```
fschuch, line 280: We need to discuss the best approach to make sure that `application_identifier` is url safe

fschuch: Further tests indicate that it is possible to create applications with complex identifiers, like:

    jobbergate applications create --name test-test --application-path ../../00_smallcase/ --identifier "Hellö Wörld@Python"

And then use this identifier to create a job script:

    jobbergate job-scripts create --name test --application-identifier "Hellö Wörld@Python"
    
The identifier is converted internally, resulting in:

    jobbergate-api_1    | INFO:     172.22.0.1:38168 - "GET /jobbergate/applications/Hell%C3%B6%20W%C3%B6rld%40Python HTTP/1.1" 200 OK

fschuch: So, I'm solving the discussion

PR #143, opened 2022-09-14
Added detailed error output when configuration errors are encountered

dusktreader:
#### What
Added configuration error details in output.

#### Why
They were hidden before, which made it impossible to see what the issue was

---

#### Peer Review


fschuch (approved):
Approved as it is, but you could also add it to the changelog if you find it convenient.

PR #148, opened 2022-09-22
Tucker/add jobbergate composed

dusktreader:
#### What
Added a subproject that wraps up all the elements of Jobbergate in a docker-compose deployment.

#### Why
This allows us to demonstrate Jobbegate as a stand-alone app, facilitates easier debugging, and serves as a good place to do end-to-end testing for Jobbergate specifically

---

#### Peer Review


examples/create-submit-job.py
```diff
@@ -30,10 +30,10 @@
 def get_application_id(
     base_api_url="https://armada-k8s.staging.omnivector.solutions/jobbergate",
```
matheushent, line 31: isn't it supposed to point to the localhost url deployed by compose?

examples/motorbike-application/templates/job-script-template.py.j2
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+#SBATCH --partition {{ data.partition }}
+#SBATCH --nodes={{ data.nodes }}
+#SBATCH --ntasks={{ data.ntasks }}
+#SBATCH -J motorbike
+#SBATCH --output=/nfs/R-%x.%j.out
+#SBATCH --error=/nfs/R-%x.%j.err
```
matheushent, line 7: make sure to add in the docs this path must exist, **and**, if run in parallel, all nodes must have access to it

jobbergate-api/CHANGELOG.rst
```diff
@@ -6,6 +6,11 @@ This file keeps track of all notable changes to jobbergate-api
 
 Unreleased
 ----------
+- Add support for supporting files on job-scripts.
+- Changed dev-tools to use alembic functions instead of subprocess calls
+- Changed default log level to DEBUG instead of INFO
+- Added ARMASEC_USE_HTTPS setting to allow non-https OIDC providers
+- Added better logging and reporting for for pydantic validation errors
```
matheushent, line 13:
```suggestion
- Added better logging and reporting for pydantic validation errors
```

jobbergate-api/CHANGELOG.rst
```diff
@@ -6,6 +6,11 @@ This file keeps track of all notable changes to jobbergate-api
 
 Unreleased
 ----------
+- Add support for supporting files on job-scripts.
```
matheushent, line 9:
```suggestion
- Added support for supporting files on job-scripts.
```

jobbergate-api/Dockerfile-dev
```diff
@@ -1,28 +1,19 @@
-FROM tiangolo/uvicorn-gunicorn-fastapi:python3.8-slim
+FROM python:3.8-slim-buster
 
-RUN apt-get -y update
-RUN apt-get install -y --fix-missing build-essential cmake libpq-dev curl
+RUN apt update && apt install -y curl libpq-dev gcc
 
-RUN pip install 'poetry==1.1.7'
+RUN curl -sSL https://install.python-poetry.org | POETRY_HOME=/opt/poetry python && \
```
matheushent, line 5: I usually fix the poetry version as well, so you can guarantee it will work as you think it will

```suggestion
RUN curl -sSL https://install.python-poetry.org | POETRY_HOME=/opt/poetry python POETRY_VERSION=1.1.13 && \
```

jobbergate-api/dev_tools/entrypoint.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
+set -e
+cd /app
 poetry run dev-tools db upgrade
-poetry run dev-tools dev-server --port=8000
-
-tail -f /dev/null
+poetry run dev-tools dev-server --port=80
```
matheushent, line 6: does it mean port 80 will be published in the host level?

jobbergate-api/jobbergate_api/main.py
```diff
@@ -85,3 +87,14 @@ async def disconnect_database():
...
+
+@app.exception_handler(RequestValidationError)
+async def validation_exception_handler(request: Request, err: RequestValidationError):
+    """
+    Handle exceptions from pydantic validators.
+    """
+    raise HTTPException(
+        status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+        detail=f"Validation error for request to {request.url} with data: {request.json()}: {err}",
+    )
```
matheushent, line 100: won't FasAPI itself do something very similar to this?

jobbergate-cli/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN curl -sSL  https://install.python-poetry.org | \
+    POETRY_HOME=/opt/poetry POETRY_VERSION=1.1.13 python && \
+    cd /usr/local/bin && \
+    ln -s /opt/poetry/bin/poetry && \
+    poetry config virtualenvs.create false
+
+COPY ./pyproject.toml ./poetry.lock* ./README* ./LICENSE* /app/
+COPY ./jobbergate_cli /app/jobbergate_cli
+WORKDIR /app
+RUN poetry install --no-dev
```
matheushent, line 14: Better to put the workdir at the beginning so you get advantages of the docker layer caching

```suggestion
FROM python:3.8-slim-buster

WORKDIR /app

RUN apt update && apt install -y curl libpq-dev gcc

RUN curl -sSL  https://install.python-poetry.org | \
    POETRY_HOME=/opt/poetry POETRY_VERSION=1.1.13 python && \
    cd /usr/local/bin && \
    ln -s /opt/poetry/bin/poetry && \
    poetry config virtualenvs.create false

COPY ./pyproject.toml ./poetry.lock* ./README* ./LICENSE* /app/
COPY ./jobbergate_cli /app/jobbergate_cli

RUN poetry install --no-dev
```

jobbergate-cli/jobbergate_cli/render.py
```diff
@@ -116,13 +116,17 @@ def render_list_results(
...
-        return
-
     if ctx.raw_output:
-        render_json(envelope.results)
+        # THIS IS SO DUMB....need to dump to string to make sure paths are converted,
+        # then convert back to dict so we can strip out the "results" element, then
+        # dump back to string
+        serialized = envelope.json()
+        deserialized = json.loads(serialized)
+        render_json(deserialized["results"])
```
matheushent, line 125: why not to use regex validation and manipulation?

jobbergate-cli/jobbergate_cli/render.py
```diff
@@ -188,7 +192,10 @@ def render_single_result(
     :param: title:         The title header to include above the ``Tale`` output
     """
     if isinstance(result, pydantic.BaseModel):
-        result = result.dict()
+        result_model = cast(pydantic.BaseModel, result)
+        result_dict = json.loads(result_model.json())
+        result = cast(Dict[str, Any], result_dict)
+
```
matheushent, line 198: why??

matheushent: The following snippet shows it is supposed to be the same thing but this implementation is not that good

```python
from pydantic import BaseModel
import json
from typing import Any, Dict, cast
import time

class Foo(BaseModel):
    foo: str

foo = Foo(foo="boo")

init = time.time()
foo_dict_easy = foo.dict()
finish = time.time()
print("Time for easy operation: {} s".format(finish - init))


init = time.time()
result_model = cast(BaseModel, foo)
result_dict = json.loads(result_model.json())
result = cast(Dict[str, Any], result_dict)
finish = time.time()
print("Time for complex operation: {} s".format(finish - init))
print()

print("Are both results equal? {}".format(foo.dict() == result))
```

jobbergate-cli/jobbergate_cli/subapps/job_submissions/tools.py
```diff
@@ -43,7 +42,6 @@ def create_job_submission(
 
     if client_id is None:
         client_id = jg_ctx.persona.identity_data.client_id
-    validate_client_id(jg_ctx, client_id)
```
matheushent, line 46: I think **if** the client is passed then it should be checked, otherwise, life goes on.

jobbergate-composed/Dockerfile-slurm
```diff
@@ -0,0 +1,156 @@
...
+
+COPY etc/slurm-env.sh /etc/profile.d/slurm-env.sh
+
+# Add local-user
+RUN useradd -ms /bin/bash local-user
+
+COPY etc/slurm-entrypoint.sh /usr/local/bin/slurm-entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/slurm-entrypoint.sh"]
+
+CMD ["slurmdbd"]
```
matheushent, line 156: this is nice stuff but too complex for this repository. I'd be more comfortable if we had a repository just to maintain this, then we can publish the image to the public and etc...

jobbergate-composed/README.rst
```diff
@@ -0,0 +1,81 @@
+=====================
+ Jobbergate Composed
+=====================
+
+_Jobbergate, Slurm, and Keycloak deployed to docker-compose._
+
+By making a simple, local deployment of Jobbergate, this project provides:
+
+* A test-bed for Jobbergate with minimal dependencies
+* An example against whih guides and tutorials can operate
```
matheushent, line 10: I think there's a typo here

examples/create-submit-job.py
dusktreader: Well, this example pre-dates it. However, I think this could use a little touch-up anyway. I will make those changes.

jobbergate-api/Dockerfile-dev
fschuch: +1 for that

jobbergate-api/dev_tools/db.py
```diff
@@ -75,16 +77,8 @@ def migrate(
...
-        "--config=alembic/alembic.ini",
-        "revision",
-        f"--message={message}",
-    ]
-    if not blank:
-        commands.append("--autogenerate")
-
-    subprocess.run(commands)
+    config = Config(file_="alembic/alembic.ini")
+    sqla_migrate(config, message=message, autogenerate=not blank)
```
fschuch, line 81: This is nice. Just to be sure, is the functionality equivalent?

examples/create-submit-job.py
dusktreader: Updated to use with jobbergate-composed.

examples/motorbike-application/templates/job-script-template.py.j2
dusktreader: This is a general problem with Slurm and Jobbergate. We don't know what paths the user will reference in their job scripts, so they can footgun themselves rather easily. We need to call it out in the docs. I will think about how to best approach that and include it in the PR for updating the Jobbergate docs.

In this case, the jobbergate-composed containers have a writable folder volume mounted at /nfs.

jobbergate-cli/jobbergate_cli/config.py
```diff
@@ -57,6 +57,7 @@ class Settings(BaseSettings):
     OIDC_DOMAIN: str
     OIDC_AUDIENCE: str
     OIDC_CLIENT_ID: str
+    OIDC_USE_HTTPS: bool = True
```
fschuch, line 60: What would be the scenarios for https and http?

fschuch, line 60: What would be the scenarios for https and http?

jobbergate-api/CHANGELOG.rst
dusktreader: Addressed in commit fe43cbd

jobbergate-composed/Dockerfile-slurm
```diff
@@ -0,0 +1,156 @@
...
+       mariadb-devel \
+       psmisc \
+       bash-completion \
+       vim-enhanced \
+       autoconf \
+       automake \
+    && yum clean all \
+    && rm -rf /var/cache/yum
+
+RUN ln -s /usr/bin/python3.4 /usr/bin/python3
```
fschuch, line 36: Why Python 3.4?

jobbergate-api/CHANGELOG.rst
dusktreader: Addressed in commit fe43cbd

jobbergate-api/dev_tools/entrypoint.sh
dusktreader: No. This entrypoint is used for containers, so their external ports will need to be forwarded to port 80 by the orchestrator.

jobbergate-api/Dockerfile-dev
dusktreader: Resolved in commit 3d7b178

fschuch (changes requested):

jobbergate-api/jobbergate_api/main.py
dusktreader: Well, you will get a 422 and a response that includes data about what is wrong with the request. However, you won't see it in the logs at all. It will just show something like "Request Failed: 422" or something like that with no details. It makes it very frustrating to debug because you have to re-issue the same request and look at the response.

jobbergate-cli/Dockerfile
dusktreader: How will setting the WORKDIR earlier in the Dockerfile help? There's nothing in workdir until the COPY command, so I don't think we would have any advantage from the layering.

jobbergate-cli/jobbergate_cli/render.py
dusktreader: JSON is a context-free grammar. It's a bad idea to parse it with regular expressions in general.
https://cstheory.stackexchange.com/a/3988

There's about a million PRs, issues, and discussions on pydantic about this issue. I'm still holding out hope that eventually they will add something like a `json_dict()` method on pydantic models for this very reason.

jobbergate-cli/jobbergate_cli/subapps/job_submissions/tools.py
dusktreader: We can't validate the client with stand-alone Jobbergate. This relies on the cluster-api, which won't be available for users that are only using Jobbergate.

We can approach this in different ways later, but for now, to allow Jobbergate to be a stand-alone platform, it needs to be removed.

jobbergate-composed/Dockerfile-slurm
dusktreader: I don't really agree. The slurm configuration and Dockerfiles found here are tailored very specifically for this use-case. It wouldn't be very useful in another context. It add anything by itself, and we'd constantly have to be pointing back to the Jobbergate main project. As it is, the users don't need to be concerned about the Docker and Slurm configuration if they are only interested in seeing how Jobbergate works.

I don't think it makes sense to publish the images either. Again, they are all targeted at this specific toy use-case. It would take A LOT more work and maintenance to make general-purpose images that would be appropriate to publish publicly, and I don't think we should tackle that problem at this point.

Having examples that can work in totality right out of the box is quite valuable.

jobbergate-composed/README.rst
dusktreader: Addressed in commit 2b65f64

jobbergate-api/dev_tools/db.py
dusktreader: Yeah, it's literally what the Alembic CLI calls directly. The difference is that with this, you don't do it in a subshell, so you can see the alembic output as it runs. Also, it's cleaner.

jobbergate-cli/jobbergate_cli/config.py
dusktreader: HTTPS should be used whenever you are hosting your APIs publicly. However, when you are using internal webservers (like local debug stacks), then it's nice to use HTTP so you don't have to worry about certs and all that mess.

So, you default to HTTPS, but allow your user to force HTTP if they ask for it.

jobbergate-composed/Dockerfile-slurm
dusktreader: Probably just because it's adapted from an old Dockerfile. We could try it with a more modern container, but it works so I'm not inclined to tinker with it much.

fschuch (approved):

jobbergate-composed/Dockerfile-slurm
matheushent: far enough mate. Resolved!

jobbergate-cli/Dockerfile
dusktreader: Addressed in commit 9b55f84

PR #151, opened 2022-09-30
ARMADA-811 Fixed sort order for enum columns

dusktreader:
#### What
Fixed sort order for enum columns.

#### Why
Because it was sorting them by insert order (of the enum)

`Task`: https://app.clickup.com/t/18022949/ARMADA-811

---

#### Peer Review

### DIFF
--- a/jobbergate-api/jobbergate_api/storage.py
+++ b/jobbergate-api/jobbergate_api/storage.py
@@ -14,7 +14,7 @@ from asyncpg.exceptions import UniqueViolationError
 from fastapi.exceptions import HTTPException
 from loguru import logger
 from sqlalchemy import Column, or_
-from sqlalchemy.sql.expression import BooleanClauseList, UnaryExpression, Case
+from sqlalchemy.sql.expression import BooleanClauseList, Case, UnaryExpression
 from starlette import status
 from yarl import URL
 
--- a/jobbergate-api/jobbergate_api/storage.py
+++ b/jobbergate-api/jobbergate_api/storage.py
@@ -65,6 +65,27 @@ def search_clause(
     return or_(*[field.ilike(f"%{term}%") for field in searchable_fields for term in search_terms.split()])
 
 
+def _build_enum_sort_clause(sort_column: Column, sort_ascending: bool) -> Case:
+    """
+    Build a Case statement that can be used as a sort clause for an enum column.
+
+    SQLAlchemy will not sort enums alphabetically by default, but rather by creation order.  Thus, we have to
+    force alphabetical sorting using a CASE clause. The logic here is strange. Basically, we have to provide a
+    lookup and then a string sort value. If sort order is ascending, each enum value's sort value is itself.
+    Otherwise, the sort value is it's opposite in the list. For example::
+
+        Given the enum (GAMMA, ALPHA, DELTA, BETA)
+        The default sort order used by SQLAlchemy is (GAMMA, ALPHA, DELTA, BETA)
+        Mapping used for ascending sort: {ALPHA: ALPHA, BETA: BETA, DELTA: DELTA, GAMMA: GAMMA}
+        Mapping used for descending sort: {ALPHA: GAMMA, BETA: DELTA, DELTA: BETA, GAMMA: ALPHA}
+
+    To understand this more fully, start with this SO question: https://stackoverflow.com/a/23618085/642511
+    """
+    sorted_values = sorted(sort_column.type.enums)
+    sort_tuple = zip(sorted_values, sorted_values if sort_ascending else reversed(sorted_values))
+    return sqlalchemy.case(dict(sort_tuple), value=sort_column)
+
+
 def sort_clause(
     sort_field: str,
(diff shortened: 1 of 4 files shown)


jobbergate-api/jobbergate_api/storage.py
```diff
@@ -81,9 +81,26 @@ def sort_clause(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=f"Invalid sorting column requested: {sort_field}. Must be one of {sort_field_names}",
         )
-    sort_column: typing.Union[Column, UnaryExpression] = sortable_fields[index]
-    if not sort_ascending:
-        sort_column = sort_column.desc()
+    sort_column: typing.Union[Column, UnaryExpression, Case] = sortable_fields[index]
+    if isinstance(sort_column.type, sqlalchemy.Enum):
+        # SQLAlchemy will not sort enums alphabetically be default, but rather by creation order.
```
matheushent, line 86: Typo

```suggestion
        # SQLAlchemy will not sort enums alphabetically by default, but rather by creation order.
```

jobbergate-api/jobbergate_api/storage.py
```diff
@@ -81,9 +81,26 @@ def sort_clause(
...
+            sort_tuple = zip(sorted_values, sorted_values)
+        else:
+            sort_tuple = zip(sorted_values, reversed(sorted_values))
+        sort_column = sqlalchemy.case(
+            {key: sort_value for (key, sort_value) in sort_tuple},
+            value=sort_column,
+        )
+    else:
+        if not sort_ascending:
+            sort_column = sort_column.desc()
```
matheushent, line 103: as you said, a bit insane 😂 Why not to use the `sorted` built-in function after the data has been fetched from the database? Definitely it would become easier to understand. About performance, well, I might be wrong, but for operations against less than a hundred thousands, it wouldn't be that bad.

Example of what I do using the `sorted` function: https://github.com/omnivector-solutions/armada-admin-api/blob/main/armada_admin_api/routers/organizations/organizations.py#L160

jobbergate-api/jobbergate_api/storage.py
```diff
@@ -81,9 +81,26 @@ def sort_clause(
...
+        # Sort value. If sort order is ascending, each enum value's sort value is itself. Otherwise,
+        # the sort value is it's opposite in the list.
+        # To begin, See: https://stackoverflow.com/a/23618085/642511
+        sorted_values = sorted(sort_column.type.enums)
+        if sort_ascending:
+            sort_tuple = zip(sorted_values, sorted_values)
+        else:
+            sort_tuple = zip(sorted_values, reversed(sorted_values))
+        sort_column = sqlalchemy.case(
+            {key: sort_value for (key, sort_value) in sort_tuple},
```
fschuch, line 98:
```suggestion
            dict(sort_tuple),
```

Optional suggestion to shorten the code.

jobbergate-api/jobbergate_api/tests/test_storage.py
```diff
@@ -53,3 +59,70 @@ def test_build_db_url__uses_TEST_prefixed_database_settings_if_passed_the_force_
         assert build_db_url() == (
             "postgresql://built-test-user:built-test-pswd@built-test-host:9999/built-test-name"
         )
+
+@pytest.mark.asyncio
+async def test_sort_clause__auto_sort_enum_column(database_engine):
```
fschuch, line 64: missing docstring

fschuch (changes requested):

jobbergate-api/jobbergate_api/storage.py
dusktreader: We could use that approach, but this is a minimal change that fixes the sorting on enum columns without disrupting the rest of the framework built around sorting that's already in place and functioning as expected.

Additionally, this is a good point of reference if I ever have to sort an enum column again! =]

jobbergate-api/jobbergate_api/storage.py
dusktreader: Nice call. Didn't even think of it.

jobbergate-api/jobbergate_api/storage.py
dusktreader: Addressed in commit af59b53

jobbergate-api/jobbergate_api/storage.py
dusktreader: Addressed in commit af59b53

jobbergate-api/jobbergate_api/tests/test_storage.py
dusktreader: Addressed in commit 103ed63

fschuch (approved):

jobbergate-api/jobbergate_api/storage.py
dusktreader: Broke this logic out into its own function with more explanation in commit 616b81b.

PR #163, opened 2022-10-11
Updated motorbike jobbergate example to use variable workdir

dusktreader:
#### What
Made the job script in the motorbike example use a template var for workdir.

#### Why
It will allow the example to run in a different environment where there may not be a `/nfs` directory to work in.

---

#### Peer Review

### DIFF
--- a/examples/motorbike-application/jobbergate.py
+++ b/examples/motorbike-application/jobbergate.py
@@ -1,5 +1,5 @@
 from jobbergate_cli.subapps.applications.application_base import JobbergateApplicationBase
-from jobbergate_cli.subapps.applications.questions import Text, Integer, Directory
+from jobbergate_cli.subapps.applications.questions import Text, Integer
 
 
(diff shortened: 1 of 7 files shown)


fschuch (approved):
Nice work. Do you mind doing the same for the example `simple-application` (or a task to do so)?

examples/motorbike-application/jobbergate.py
```diff
@@ -1,5 +1,5 @@
 from jobbergate_cli.subapps.applications.application_base import JobbergateApplicationBase
-from jobbergate_cli.subapps.applications.questions import Text, Integer
+from jobbergate_cli.subapps.applications.questions import Text, Integer, Directory
```
matheushent: you're not actually using this new import, are you?

matheushent (approved):

examples/motorbike-application/jobbergate.py
dusktreader: resolved in latest commit.

PR #189, opened 2022-11-25
ASP-2254 Add "execution parameters" to Job Submissions

fschuch:
#### What
* Add a column to the job_submissions table named `execution_parameters`
* Add support for `execution_parameters` as part of the POST endpoint for `job_submissions`
* `execution_parameters` should not be changeable through any PUT or PATCH endpoints (once submitted, you should not be able to change these)

#### Why
We need to move where SBATCH params are designated so that they are attached to particular Job Submissions instead of keeping them with Job Scripts (as a part of the source file).

`Task`: https://jira.scania.com/browse/ASP-2254

---

#### Peer Review


jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -104,7 +104,7 @@ async def job_submission_create(
 
     logger.debug(f"Job-submission created: {job_submission_data=}")
 
-    return job_submission_data
+    return JobSubmissionResponse(**job_submission_data)
```
fschuch, line 107: This line needs more attention. Returning just `job_submission_data`, as in the original code, results in a Pydantic validation error, which is just weird because we can create the model in the new code with no problems. 

Mypy is complaining about the new code since the argument after ** must be a mapping.

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
```diff
@@ -60,9 +60,218 @@
...
+    cpu_binding: Optional[str] = Field(description="Cpu binding")
+    cpu_binding_hint: Optional[str] = Field(description="Cpu binding hint")
+    cpu_frequency: Optional[str] = Field(
+        description=(
+            "Request that job steps initiated by srun commands inside this sbatch "
+            "script be run at some requested frequency if possible, on the CPUs "
+            "selected for the step on the compute node(s)."
+        ),
+    )
+    cpus_per_gpu: Optional[str] = Field(description="Number of CPUs requested per allocated GPU.")
```
matheushent, line 125: you sure this is a string?

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
```diff
@@ -60,9 +60,218 @@
...
+    qos: Optional[str] = Field(description="Request a quality of service for the job.")
+    requeue: Optional[bool] = Field(
+        description="Specifies that the batch job should eligible to being requeue."
+    )
+    reservation: Optional[str] = Field(
+        description="Allocate resources for the job from the named reservation."
+    )
+    signal: Optional[str] = Field(
+        description="When a job is within sig_time seconds of its end time, send it the signal sig_num.",
+        # regex=r"[B:]<sig_num>[@<sig_time>]",
```
matheushent, line 225: leftover?

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
fschuch: According to slurm's documentation, it is:
image

It is also supported by the slurm API's JSON schema:

```
"cpus_per_gpu": {
    "type": "string",
    "description": "Number of CPUs requested per allocated GPU."
},
```

Thanks for pointing it out, I agree `int` would make more sense, I'll keep an eye on this when working on the next steps of this task. This was the first sub-task, and I need to work on a couple more to be able to perform integration tests.

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
fschuch: This is actually a placeholder for some further investigation. According to the documentation, this parameter has to follow a pattern:
```
"signal": {
    "pattern": "[B:]<sig_num>[@<sig_time>]",
    "type": "string",
    "description": "When a job is within sig_time seconds of its end time, send it the signal sig_num."
},
```

I'm wondering if we can check the pattern using some regex to inform Jobbergate's users about the error earlier in the workflow. As I said in the comment above, this was the first sub-task, and I need to work on a couple more to be able to perform integration tests and check it.

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
```diff
@@ -60,9 +60,218 @@
...
+        description="Specify alternate distribution methods for remote processes."
+    )
+    environment: Optional[Dict[str, str]] = Field(description="Dictionary of environment entries.")
+    exclusive: Optional[Literal["user", "mcs", "exclusive", "oversubscribe"]] = Field(
+        description=(
+            "The job allocation can share nodes just other users with the "
+            "'user' option or with the 'mcs' option)."
+        )
+    )
+    get_user_environment: Optional[bool] = Field(
```
fschuch, line 161: This one is `int` the cluster-agent, another point to keep an eye on integration tests:
https://github.com/omnivector-solutions/cluster-agent/blob/229b35ca804c60831ba3c9ba6ba4b7253773def4/cluster_agent/jobbergate/schemas.py#L46

matheushent: indeed is an int. I checked my internal notes and I verified it is an int. `1` or `0`.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
fschuch: solved in [60cea43](https://github.com/omnivector-solutions/jobbergate/pull/189/commits/60cea43c3b15c7a83086f7bd5e4683e8915f3359)

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -104,7 +104,7 @@ async def job_submission_create(
 
     logger.debug(f"Job-submission created: {job_submission_data=}")
 
-    return job_submission_data
+    return JobSubmissionResponse(**job_submission_data)  # type: ignore
```
dusktreader, line 107: Why does this require an ignore annotation?

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
```diff
@@ -60,9 +60,218 @@
         description="The report message received from cluster-agent when a job submission is rejected",
         example="Unrecognized SBATCH arguments",
     ),
+    execution_parameters=MetaField(
+        description="The parameters to be passed to the job submission",
```
dusktreader, line 64: Can you please include a link to the Slurm documentation for accepted parameters?

dusktreader (approved):

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
matheushent: we've been discussing this privately on: https://omnivector.slack.com/archives/C03FP13MAE6/p1669737235008909

dusktreader: The issue is that the type of `job_submission_data` is not a dict. This can be fixed by using the `from_orm()` method that initializes the response object from another object:

```python
return JobSumbissionResponse.from_orm(job_submission_data)
```

Mypy accepts this just fine.

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
fschuch: solved in [5c44cbe](https://github.com/omnivector-solutions/jobbergate/pull/189/commits/5c44cbe56ba5551c97cd0c1d9ae7739980153270)

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
fschuch: Mypy accepts it, but pydantic does not, there is an issue related to this: https://github.com/tiangolo/pydantic-sqlalchemy/issues/19

I'm leaving the code as it is for now.

jobbergate-api/jobbergate_api/apps/job_submissions/schemas.py
fschuch: solved in [235255a](https://github.com/omnivector-solutions/jobbergate/pull/189/commits/235255af07fb2a145bbcdcdf49abb795643a2b65)

fschuch:
> So, general comment: don't you think it would be good to have tests that check all execution params? I know it's quite of a work but I won't be safe approving this without it.

After discussing the point, we agreed that there is no need to test the schema itself, but the route and database connections that use the information. This is addressed in another PR: https://github.com/omnivector-solutions/jobbergate/pull/193

PR #193, opened 2022-12-02
ASP-2256 Migrate the SBATCH param mapper to jobbergate-api

fschuch:
to the `job_submissions`  POST endpoint

#### What

The sbatch parser was implemented successfully on the cluster-agent. 

The implementation worked, but it was difficult to reflect errors to the user due to the asynchronous nature of remote job submission. Raising these errors early in the workflow will result in a better user experience, in general. Jobbergate would also present the error message and log, allowing users to act and solve the problems related to the parameters.

#### Why
Explain why the proposed change is necessary.

`Task`: https://jira.scania.com/browse/ASP-2256

---

#### Peer Review


jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
```diff
@@ -199,8 +211,12 @@ async def test_create_job_submission__with_execution_directory(
...
     create_data.pop("client_id", None)
 
-    with time_frame() as window:
-        response = await client.post("/jobbergate/job-submissions/", json=create_data)
+    with mock.patch(
+        "jobbergate_api.apps.job_submissions.routers.get_job_properties_from_job_script"
+    ) as mocked:
+        mocked.return_value = JobProperties.parse_obj(create_data["execution_parameters"])
+        with time_frame() as window:
+            response = await client.post("/jobbergate/job-submissions/", json=create_data)
```
matheushent, line 219: why did you prefer to mock instead of "prepare" the test by adding a job script to the database previously?

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
```diff
@@ -117,20 +122,27 @@ async def test_create_job_submission__with_client_id_in_request_body(
...
-        )
+
+    execution_parameters = {
+        "name": "job-submission-name",
+        "comment": "I am a comment",
+    }
+
+    with mock.patch(
+        "jobbergate_api.apps.job_submissions.routers.get_job_properties_from_job_script"
+    ) as mocked:
```
matheushent, line 133: same question applies here

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_properties_parser.py
```diff
@@ -0,0 +1,657 @@
...
+    actual_value = _clean_line(line)
+    assert actual_value == desired_value
+
+
+@pytest.mark.parametrize(
+    "line, desired_value",
+    (
+        ("--help", ["--help"]),
+        ("--abc=0", ["--abc", "0"]),
+        ("--abc = 0", ["--abc", "0"]),
```
matheushent, line 91: this one shouldn't be even acceptable since slurm rejects it. See my screenshot:

image

Tested on Ubuntu 20.04. Deployed by the slurm charms

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -90,6 +91,16 @@ async def job_submission_create(
             detail=message,
         )
 
+    try:
+        job_properties = get_job_properties_from_job_script(
+            job_submission.job_script_id, **job_submission.execution_parameters.dict(exclude_unset=True)
+        )
+        create_dict["execution_parameters"] = job_properties.dict(exclude_unset=True)
```
matheushent, line 98: I know the variable creation comes from other place but let's agree a variable named `create_dict` isn't that descriptive. It indicates an action, quite weird I'd say.

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
fschuch: It was preferred following the encapsulation practice. The route has no need to know the implementation details from `get_job_properties_from_job_script` and neither does its test. In addition, there is no need to change this test if these internal implementation details change.

matheushent: hmmmmm, I don't think this is a good approach tbh. Let me think in some solid points then I update this thread

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
```diff
@@ -80,6 +86,10 @@ async def test_create_job_submission__with_client_id_in_token(
     assert job_submission.created_at in window
     assert job_submission.updated_at in window
 
+    assert job_submission.execution_parameters is not None
+    assert job_submission.execution_parameters.name == "job-submission-name"
+    assert job_submission.execution_parameters.comment == "I am a comment"
```
matheushent, line 91: it is missing to check in the database if the data is indeed there. If you don't check that, it is possible that the data isn't there but the response is the one you expect.

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
```diff
@@ -142,6 +163,10 @@ async def test_create_job_submission__with_client_id_in_request_body(
...
```
matheushent, line 168: same comment apply about checking the database apply here. If it is not clear what I mean, check this references:

[1. Check if data supposedly inserted in elasticsearch is indeed there](https://github.com/omnivector-solutions/armada-api/blob/main/tests/routers/test_agent.py#L134-L141)
[2. Check if data inserted in postgres is indeed there](https://github.com/omnivector-solutions/armada-api/blob/main/tests/routers/test_graphql.py#L276-L282)

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
```diff
@@ -208,6 +238,10 @@ async def test_create_job_submission__with_execution_directory(
...
```
matheushent, line 243: again

matheushent (changes requested):

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
matheushent: so, you mock but at the end you don't assert if the mock has been actually called. Two things:

1. If you check the mock then you're testing something that shouldn't matter (because I can implement the same behaviour in a different way)
2. If you don't check the mock the it means you don't need to mock, right?

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
matheushent: in this one the code actually call `database.fetch_all`, so it is just a case of improving a little bit to check if the data in the database is exactly what we want to.

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
fschuch: That is a good catch, the assertion was missing, it was added in [2a11835](https://github.com/omnivector-solutions/jobbergate/pull/193/commits/2a11835048b6bfa416ec65d469d1733f5464cb2a).

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_properties_parser.py
fschuch: fixed in [cfb1378](https://github.com/omnivector-solutions/jobbergate/pull/193/commits/cfb13783121304f9574f47be0ae4d859cb23aa1b)

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
fschuch: solved in [cb83d36](https://github.com/omnivector-solutions/jobbergate/pull/193/commits/cb83d36dd346db4174308fad60a9f9480cae0e75)

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
matheushent: Here it is an example of what I have trying to mean:

```python
from datetime import datetime

@freeze("2022-12-05")
async def dummy_test():

    # manually insert here the application data
    # manually insert here the job script data

    payload = {
        "job_submission_name": "dummy",
        "job_submission_description": "dummy",
        "job_script_id": 1,
        "execution_directory": "/tmp",
        "client_id": "dummy",
        "execution_parameters": {
            "account": "dummy",
            "argv": "arg1 arg2 arg3 arg4",
            "environment": {"PATH": "/bin:/usr/bin/:/usr/local/bin/"},
            "partition": "compute",
        }
    }

    response = await client.post("/jobbergate/job-submissions/", json=payload)

    # assert whatever you want here

    # check if data is correctly in the database

    query = table.select().select(table.c.id == 1)
    record = await database.fetch_one(query)

    assert record.get("id") == 1 # this is the job submission ID. Check if it is 1 because the database is supposed to be empty
    assert record.get("job_submission_name") == "dummy"
    assert record.get("job_submission_description") == "dummy"
    assert record.get("job_submission_owner_email") == "whatever"
    assert record.get("job_script_id") == 1
    assert record.get("execution_directory") == "/tmp"
    assert record.get("slurm_job_id") is None
    assert record.get("client_id") == "dummy"
    assert record.get("status") == "CREATED"
    assert record.get("report_message") == str()
    assert record.get("created_at") == str(datetime.now())
    assert record.get("updated_at") == str(datetime.now())
    assert record.get("execution_parameters") == {
        "account": "dummy",
        "argv": "arg1 arg2 arg3 arg4",
        "environment": {"PATH": "/bin:/usr/bin/:/usr/local/bin/"},
        "partition": "compute",
    }
```

In this example you would really test if the endpoint works as expected: you input some data and expect this data to be in the database. Note that the code is agnostic on how you do it, so you could do by using the parser you built or whatever. There's no need to mock data.

matheushent: Note also that, there's no need to check if the variable `record` is None because, if it is, the test will fail and that's ok, it should fail indeed

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
fschuch: It was addressed in https://github.com/omnivector-solutions/jobbergate/pull/193#discussion_r1041057788

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_properties_parser.py
```diff
@@ -0,0 +1,656 @@
...
+    "line, desired_value",
+    (
+        ("--help", ["--help"]),
+        ("--abc=0", ["--abc", "0"]),
+        ("-J job_name", ["-J", "job_name"]),
+        ("-v -J job_name", ["-v", "-J", "job_name"]),
+        ("-J job_name -v", ["-J", "job_name", "-v"]),
+        ("-a=0", ["-a", "0"]),
+        ("-a = 0", ["-a", "0"]),
+        ("-a 0", ["-a", "0"]),
```
matheushent, line 96: so, I was running some tests locally and I noticed slurm only accepts the `=` sign when using flags with `--`. When the flag is set with `-`, slurm accepts a white space between the argument and the value. 

In case you try to pass `=` when you should be using white space, slurm will point out an error. Example
image

I think we should perhaps adjust this parser to only accept what slurm accepts. Thoughts?

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
```diff
@@ -58,19 +59,32 @@ async def test_create_job_submission__with_client_id_in_token(
...
+    with mock.patch(
+        "jobbergate_api.apps.job_submissions.routers.get_job_properties_from_job_script"
+    ) as mocked:
+        mocked.return_value = JobProperties.parse_obj(create_data["execution_parameters"])
+        with time_frame() as window:
+            response = await client.post("/jobbergate/job-submissions/", json=create_data)
+        mocked.assert_called_once_with(
+            inserted_job_script_id,
+            **create_data["execution_parameters"],
+        )
```
matheushent, line 71: still that conversation about mocking this. Is this really needed? If so, why? Because you're fixing how the execution parameters are set, it disables any other developer (even yourself) of doing the same thing but with a different approach.

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_properties_parser.py
matheushent: don't you think it is inconsistent to accept parameters that slurm itself doesn't?

It can drive users to run jobs by jobbergate that won't be run by slurm. Thinking as an engineer who might want to use jobbergate, I think it is very inconsistent and strange such a behaviour

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
fschuch: `get_job_properties_from_job_script` is mocked because internally it obtains the job script file from s3 in order to extract the SBATCH parameters. Here the choice was to mock just the high-level function since `get_job_properties_from_job_script` has its own tests mocking low-level functions on another test file.

matheushent: I see. Instead of mocking, have you thought about instantiating a minio server during the tests? This is something I do [here](https://github.com/omnivector-solutions/armada-api/blob/main/docker-compose.yaml#L30-L86) to test against real server. I know such an implementation is quite far from the scope of the PR but it doesn't seem ideal to me the way it is. What do you think? I could help to implement it in parallel why your work on this PR

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_properties_parser.py
fschuch: Ok, I agree it should be investigated in fixed, but not on this PR since the syntax as expected by `srun` is already covered. In fact, the syntax here should not be more permissive than that on slurm. I updated the Technical notes on [PFT-631](https://app.clickup.com/t/18022949/PFT-631) and moved it to Priority Backlog.

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
fschuch: It is a good point, but I agree in parts. Unit tests should test individual components and run as fast as possible to provide us with quick feedback, making the necessary setup and request to external services does not help with that, so I'm in favor of mocking them.

Integration tests, on the other hand, are a good place for that and I agree Jobbergate could make good use of them. Your example is a good use of the concept, and I have another one I've worked with [here](https://github.com/fschuch/file_storehouse/blob/main/tests/test_user_story.py). I'm happy to have a new task covering this topic.

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_properties_parser.py
matheushent: perfect, thank you

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
matheushent: > making the necessary setup and request to external services do not help with that,

Quite the opposite. Testing against a real database will give you the exact behaviour you'd have in real deployments.

About the speed: jobbergate tests are slower than the tests I sent you... I still can't see how using a database for testing cannot be beneficial over than the way it currently is.

As I said, this implementation is far from the scope of the task itself but definitely it can be improved. I'm not going to stop your development but I'd much appreciate if we focus on improving that in the near future, and the reason is simple: the way it is doesn't test real scenarios. It only tests if you called a function or not but the implementation is possible in countless ways.

fschuch: I appreciate the comments, I have no doubt we can converge on a solution in the near future, but at the same time, I think we agree it is out of the scope of this PR.

matheushent: do you mind writting a task for that?

matheushent (approved):

PR #195, opened 2022-12-07
ASP-2300 Add "execution_parameters" at the CLI

fschuch:
#### What
Add support at the command line interface for the new feature "execution_parameters" for a job submission

#### Why
Explain why the proposed change is necessary.

`Task`: https://jira.scania.com/browse/ASP-2300

---

#### Peer Review

### DIFF
--- a/jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
+++ b/jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
@@ -144,7 +144,7 @@ async def job_submission_get(job_submission_id: int = Query(...)):
 
     logger.debug(f"Job-submission data: {job_submission_data=}")
 
-    return job_submission_data
+    return JobSubmissionResponse(**job_submission_data)  # type: ignore
 
 
 @router.get(
--- a/jobbergate-cli/jobbergate_cli/subapps/job_submissions/app.py
+++ b/jobbergate-cli/jobbergate_cli/subapps/job_submissions/app.py
@@ -3,6 +3,7 @@ Provide a ``typer`` app that can interact with Job Submission data in a cruddy m
 """
 
 from pathlib import Path
+from textwrap import dedent
 from typing import Any, Dict, Optional, cast
 
 import typer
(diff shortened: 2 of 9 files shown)


jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -144,7 +144,7 @@ async def job_submission_get(job_submission_id: int = Query(...)):
 
     logger.debug(f"Job-submission data: {job_submission_data=}")
 
-    return job_submission_data
+    return JobSubmissionResponse(**job_submission_data)  # type: ignore
```
dusktreader, line 147: I would prefer to find a better way to do this. We shouldn't be adding ignores for types if we can avoid it.

We can look at explicitly constructing a dictionary from the result rows a la sqlalchemy Row methods: https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Row

It would also be possible to set `orm_mode` for the JobSubmissionResponse model and then use the `from_orm()` method to initialize it with the result rows themselves.

jobbergate-cli/jobbergate_cli/subapps/job_submissions/app.py
```diff
@@ -50,12 +52,26 @@ def create(
...
             If provided as a relative path, it will be converted as an absolute path from your current
-            working directory. If you use "~" to denote your home directory, the path will be expanded to an absolute
-            path for your home directory on *this* machine.
-        """,
+            working directory. If you use "~" to denote your home directory, the path will be expanded to an
+            absolute path for your home directory on *this* machine.
+            """
+        ).strip(),
+    ),
+    execution_parameters: Optional[Path] = typer.Option(
```
dusktreader, line 64: I'm not so sure about this requiring a JSON file path. Why did you decide to go this route?

dusktreader (changes requested):

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
fschuch: I've tried `from_orm()`, but it results in an error when we have any data on the column `execution_parameters` (JSONB)

jobbergate-cli/jobbergate_cli/subapps/job_submissions/app.py
dusktreader: Oh, I see. Ok.

dusktreader (approved):

PR #215, opened 2023-01-13
Patch optional field execution_parameters when creating job-submissions

fschuch:
#### What
It was not possible to create job-submissions from the UI. The last release included the new field `execution_parameters` that should be optional, but an error is raised when it is not on the request.

#### Why
The issue was identified by the front-end in staging after Jobbergate-API was updated to 3.4.0. Notice the issue does not affected the usability from the CLI.

---

#### Peer Review

### DIFF
--- a/jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
+++ b/jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
@@ -137,10 +137,11 @@ async def test_create_job_submission__without_execution_parameters(
         values=fill_job_script_data(application_id=inserted_application_id),
     )
 
+    dummy_client_id = "dummy-cluster-client"
     inject_security_header(
         "owner1@org.com",
         Permissions.JOB_SUBMISSIONS_EDIT,
-        client_id="dummy-cluster-client",
+        client_id=dummy_client_id,
     )
     create_data = fill_job_submission_data(
         job_script_id=inserted_job_script_id,
@@ -179,12 +180,16 @@ async def test_create_job_submission__without_execution_parameters(
 
     # Check that each field is correctly set
     assert job_submission.id == job_submission_raw_data.get("id")
-    assert job_submission.job_submission_name == "sub1"
-    assert job_submission.job_submission_owner_email == "owner1@org.com"
-    assert job_submission.job_submission_description is None
-    assert job_submission.job_script_id == inserted_job_script_id
-    assert job_submission.execution_directory is None
-    assert job_submission.client_id == "dummy-cluster-client"
+    assert job_submission.job_submission_name == create_data.get("job_submission_name")
+    assert job_submission.job_submission_owner_email == create_data.get("job_submission_owner_email")
+    assert job_submission.job_submission_description == create_data.get("job_submission_description")
+    assert job_submission.job_script_id == create_data.get("job_script_id")
+    assert job_submission.execution_directory == create_data.get("execution_directory")
+
+    # client_id was not on the payload, it should be set by other mechanisms
+    assert create_data.get("client_id") is None
+    assert job_submission.client_id == dummy_client_id
+
     assert job_submission.status == JobSubmissionStatus.CREATED
     assert job_submission.created_at in window
     assert job_submission.updated_at in window
--- a/jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
+++ b/jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
@@ -109,6 +109,90 @@ async def test_create_job_submission__with_client_id_in_token(
 
 
 @pytest.mark.asyncio
+async def test_create_job_submission__without_execution_parameters(
+    fill_application_data,
+    fill_job_script_data,
+    fill_job_submission_data,
+    client,
+    inject_security_header,
+    time_frame,
+):
+    """
+    Test POST /job-submissions/ correctly creates a job_submission.
+
+    This test proves that a job_submission is successfully created via a POST request to the /job-submissions/
+    endpoint. We show this by asserting that the job_submission is created in the database after the post
+    request is made, the correct status code (201) is returned. We also show that the ``client_id``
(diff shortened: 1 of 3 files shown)


jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
```diff
@@ -108,6 +108,90 @@ async def test_create_job_submission__with_client_id_in_token(
...
+    # Removed defaults to make sure these are correctly set by other mechanisms
+    create_data.pop("status", None)
+    create_data.pop("client_id", None)
+
+    with mock.patch(
+        "jobbergate_api.apps.job_submissions.routers.get_job_properties_from_job_script"
+    ) as mocked:
+        mocked.return_value = JobProperties()
+        with time_frame() as window:
+            response = await client.post("/jobbergate/job-submissions/", json=create_data)
```
matheushent, line 160: part of me feels weird to keep this forward slash. I can't remember exactly what happened but I remember me and @dusktreader being stuck for a long in Mexico trying to find out a problem and at the end we discovered the problem was a damn forward slash. Do you mind removing it and testing as our frontend and the jobbergate cli uses

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
```diff
@@ -108,6 +108,90 @@ async def test_create_job_submission__with_client_id_in_token(
...
+    assert job_submission.id == job_submission_raw_data.get("id")
+    assert job_submission.job_submission_name == "sub1"
+    assert job_submission.job_submission_owner_email == "owner1@org.com"
+    assert job_submission.job_submission_description is None
+    assert job_submission.job_script_id == inserted_job_script_id
+    assert job_submission.execution_directory is None
+    assert job_submission.client_id == "dummy-cluster-client"
+    assert job_submission.status == JobSubmissionStatus.CREATED
+    assert job_submission.created_at in window
+    assert job_submission.updated_at in window
```
matheushent, line 190: so you're checking if the job submission data returned in the response is the one that is in the database. Despite it kind of making sense, it is possible the test is not testing the reality. Check the following piece of code:

```python
@app.post("/whatever")
async def whatever(body: Model):
    # build a row based on nothing and completely ignore the input model
    body = Model(jobbergate="is cool", testing="is not cool")

    # logic to insert the body model in the database
```

In this code I showed you that your test would succeed but the data in the database isn't what you really want.

So, to fix the test is simple: **you need to check the database data with the `create_data` variable**.

Does it make sense to you?

matheushent (changes requested):
Nice quick fix, I liked. I just have a point for discussion related to the test.

Also, we agreed in the `jobbergate-discussion` channel this would be a patch in the 3.4.1 version. However, looking at the changelog, I see more stuff. Do you have plans to put all the changelog changes in 3.4.1? If so, I think it is better to make it clear in the `jobbergate-discussion` channel.

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
fschuch: The test passes without the forward slash `response = await client.post("/jobbergate/job-submissions", json=create_data)`.

Almost all tests on the routes use it, so the investigation and removal of them may be out of scope for this PR. I agree that is a valid point to keep in mind.

fschuch:
> Nice quick fix, I liked. I just have a point for discussion related to the test.
> 
> Also, we agreed in the `jobbergate-discussion` channel this would be a patch in the 3.4.1 version. However, looking at the changelog, I see more stuff. Do you have plans to put all the changelog changes in 3.4.1? If so, I think it is better to make it clear in the `jobbergate-discussion` channel.

Usability changes would require further discussions, but that is not the case since all changes are patching minor issues. I see no reason not to include them in the next patch release. I'll make it clear on the channel as you suggested.

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
matheushent: > On the other hand, on lines 181-190, we make sure that each field has the expected value based on the test setup, which confirms that the data on both response and database is what we want

In line 168 you just select a random job submisson from the database. You **expect** it to be the one created by the endpoint but you cannot guarantee it is the same unless you check against the `create_data` variable. The `create_data` variable defined in line 145 is only used as payload for the request, by my snippet precisely showed you that you can completely ignore the payload, create an entry in the database and then make the test a fool test.

fschuch: Thanks for your suggestion, it was addressed on [ba92d0b](https://github.com/omnivector-solutions/jobbergate/pull/215/commits/ba92d0b54fe5114cc0a5ecc54e2db76f90c01669)

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
```diff
@@ -179,12 +180,16 @@ async def test_create_job_submission__without_execution_parameters(
...
-    assert job_submission.execution_directory is None
-    assert job_submission.client_id == "dummy-cluster-client"
+    assert job_submission.job_submission_name == create_data.get("job_submission_name")
+    assert job_submission.job_submission_owner_email == create_data.get("job_submission_owner_email")
+    assert job_submission.job_submission_description == create_data.get("job_submission_description")
+    assert job_submission.job_script_id == create_data.get("job_script_id")
+    assert job_submission.execution_directory == create_data.get("execution_directory")
+
+    # client_id was not on the payload, it should be set by other mechanisms
+    assert create_data.get("client_id") is None
```
matheushent, line 190: how should it be set then?

matheushent (approved):

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_routers.py
fschuch: it comes from the token

PR #222, opened 2023-01-27
ASP-2622 Handle special characters in the Jobbergate SBATCH parser

fschuch:
#### What
The SBATCH parser does not support any value that includes white spaces or the equal character `=`.

#### Why
As a nextgen Jobbergate user, I need the SBATCH parser in the API to handle special characters correctly so that it can handle all the different SBATCH parameters that are currently being used in the legacy use-case.

Investigating the SBATCH parameters, two parameters were identified that are supposed to use the special characters (--acctg-freq  and --comment ), for instance:

```python
#SBATCH --acctg-freq=task=60,energy=30,network=30,filesystem=60
#SBATCH --comment="Hi, I am a comment"
```

`Task`: https://jira.scania.com/browse/ASP-2622

---

#### Peer Review

### DIFF
--- a/jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_properties_parser.py
+++ b/jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_properties_parser.py
@@ -20,7 +20,6 @@ from jobbergate_api.apps.job_submissions.properties_parser import (
     _clean_jobscript,
     _clean_line,
     _flagged_line,
-    _split_line,
     build_mapping_sbatch_to_slurm,
     build_parser,
     convert_sbatch_to_slurm_api,
@@ -28,6 +27,7 @@ from jobbergate_api.apps.job_submissions.properties_parser import (
     get_job_properties_from_job_script,
     jobscript_to_dict,
     sbatch_to_slurm_mapping,
+    split,
 )
 from jobbergate_api.apps.job_submissions.schemas import JobProperties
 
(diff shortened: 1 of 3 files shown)


julianaklulo (approved):
Nice!

jobbergate-api/jobbergate_api/tests/apps/job_submissions/test_properties_parser.py
```diff
@@ -91,22 +91,20 @@ def test_clean_line(line, desired_value):
...
     (
         ("--help", ["--help"]),
-        ("--abc=0", ["--abc", "0"]),
+        ("--abc=0", ["--abc=0"]),
         ("-J job_name", ["-J", "job_name"]),
         ("-v -J job_name", ["-v", "-J", "job_name"]),
         ("-J job_name -v", ["-J", "job_name", "-v"]),
-        ("-a=0", ["-a", "0"]),
-        ("-a = 0", ["-a", "0"]),
+        ("-a=0", ["-a=0"]),
```
matheushent, line 98: the slurm version supported by the charms doesn't accept the equal operator with single character parameters. For example, consider the following command:

```bash
srun -w=juju-7561f2-8,juju-7561f2-11 -p osd-slurmd hostname
```

It raises the following error:

```
srun: error: Unable to allocate resources: Invalid node name specified
```

That said, I wonder how useful is to test something that won't be used actually.

fschuch: These are base checks for `split` from the build-it module `shlex`. As stated in the test's docstring, it just aims to record how the function behaves.

matheushent: isn't it redundant them? I suppose the function is already quite well tested when published with Python. The test is named `test_properties_parser`, so I expect it to test the slurmrestd job properties payload, not a Python function behaviour.

fschuch: I agree it is tested since it is built-in in Python, that is not the intention here. This test is just a way to record or document briefly what I've learned about `shlex.split` on this task, it will be available as a reference together with all other components of the property parser.

PR #228, opened 2023-02-16
ASP-2810 Create auth helpers in Jobbergate-core

fschuch:
#### What

* A package called "jobbergate-core" is published on pypi
* The package contains methods to fetch and refresh access tokens
* The source code is a part of the main Jobbergate git project
* Documentation is provided on how to use it

#### Why
As a simulation engineer working on automation, I want a python package that can help me secure an API key to make Jobbergate API requests.

`Task`: https://jira.scania.com/browse/ASP-2810

---

#### Peer Review


jobbergate-core/README.rst
```diff
@@ -0,0 +1,15 @@
+=================
+ Jobbergate Core
+=================
+
+Jobbergate-core is a sub-project that contains the key components and logic that is shared among all others sub-projects (CLI, API, and Agent), and that can be accessed by the end-users in order to support their custom automation built on top of Jobbergate.
```
jamesbeedy, line 5:
```suggestion
Jobbergate-core is a sub-project that contains the key components and logic that is shared among all other sub-projects (CLI, API, and Agent), and that can be accessed by the end-users in order to support their custom automation built on top of Jobbergate.
```

jobbergate-core/CHANGELOG.rst
```diff
@@ -0,0 +1,8 @@
+============
+ Change Log
+============
+
+This file keeps track of all notable changes to jobbergate-core
+
+Unreleased
+----------
```
fschuch, line 8: Update the changelog with what was added in this PR

jobbergate-core/jobbergate_core/version.py
```diff
@@ -0,0 +1,9 @@
+"""
+Provide the version of the package.
+"""
+try:
+    from importlib import metadata  # type: ignore
+except ImportError:
+    import importlib_metadata as metadata  # type: ignore # Python < 3.8
+
+__version__ = metadata.version(__package__ or __name__)
```
matheushent, line 9: I just see this variable is being used during tests. What's the purpose and utility of it?

jobbergate-core/jobbergate_core/auth/token.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+
+    ACCESS = "access"
+    REFRESH = "refresh"
+
+
+@dataclass(frozen=True)
+class Token:
+    """
+    Class for handling tokens.
```
fschuch, line 32: Document the parameters necessary to create this object.

jobbergate-core/jobbergate_core/version.py
```diff
@@ -0,0 +1,9 @@
+"""
+Provide the version of the package.
+"""
+try:
+    from importlib import metadata  # type: ignore
```
fschuch, line 5: It can be removed once the requirements were updated to Python 3.8 of higher

jobbergate-docs/src/core.rst
```diff
@@ -0,0 +1,24 @@
+=================
+ Jobbergate Core
+=================
+
```
matheushent, line 4: I feel a little bit of lack of explanation here. See the image below:
image

To understand a little bit better what is the jobbergate-core I need to go to the source code and read the readme. If this is the flow, I think the doc has failed. Maybe a simple inception here improves the doc :)

jobbergate-docs/src/conf.py
```diff
@@ -30,11 +30,17 @@
...
     "sphinx.ext.autosectionlabel",
+    "sphinx.ext.githubpages",
+    "sphinx.ext.intersphinx",
+    "sphinx.ext.napoleon",
     "sphinxcontrib.httpdomain",
 ]
 
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3", None,),
+}
```
fschuch, line 43: Check if `requests` and/or `httpx` can be added here, to make them easier to refer.

jobbergate-core/jobbergate_core/version.py
fschuch: The version is available as metadata for end users that installed it from Pypi, for instance:

```python
>>> import jobbergate_core
>>> print(jobbergate_core.__version__)
3.4.3
```

It is a small touch, be can be useful sometimes.

fschuch: The reference came from here: https://stackoverflow.com/questions/67085041/how-to-specify-version-in-only-one-place-when-using-pyproject-toml

jobbergate-docs/src/core.rst
fschuch: Fair point, I'll elaborate

jobbergate-docs/src/conf.py
```diff
@@ -30,11 +30,17 @@
...
 exclude_patterns = []
 extensions = [
-    "sphinx.ext.githubpages",
+    "sphinx.ext.autodoc",
     "sphinx.ext.autosectionlabel",
+    "sphinx.ext.githubpages",
+    "sphinx.ext.intersphinx",
+    "sphinx.ext.napoleon",
     "sphinxcontrib.httpdomain",
 ]
```
matheushent, line 39: a huge addition here IMO is `sphinx_copybutton`. The code snippets can be coppied to clipboard quite easily. Check the links bellow for details:

[1] https://github.com/omnivector-solutions/armada-infrastructure/blob/main/docs-source/conf.py#L41
[2] https://github.com/omnivector-solutions/armada-infrastructure/blob/main/pyproject.toml#L17

jobbergate-docs/src/core.rst
fschuch: Have you built it locally to review? Thanks for this level of attention to detail.

jobbergate-docs/src/conf.py
fschuch: Totally

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+@dataclass
+class JobbergateAuth:
+    """
+    High-level class used to handle authentication in Jobbergate.
+
+    After an instance of this class is created, it can be used to authenticate requests
+    from the `requests`_ library or the `httpx`_ library by passing it to the ``auth``
+    parameter on the request (see examples bellow).
```
matheushent, line 30:
```suggestion
    After an instance of this class is created, it can be used to authenticate requests
    from both `requests`_  and `httpx`_ packages by passing it to the ``auth``
    parameter on the request (see examples bellow).
```

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+
+    Arguments:
+        cache_directory (Path): Directory to be used for the caching tokens.
+        login_domain (str): Domain used for the login.
+        login_audience (str): Audience of the login.
+        login_client_id (str): Client ID used for login.
+        tokens (Dict[TokenType, Token]): Dictionary holding the tokens needed for authentication.
+
+    Note:
+        Consult the values above with your system administrator.
```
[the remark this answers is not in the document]
fschuch: I've written it more generically having in mind it is open-source documentation, and Jobbergate may be running on systems we do not have control.

At the same time, I see your point. Thoughts @matheushent @dusktreader @jamesbeedy?

jobbergate-core/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	@rm -rf .mypy_cache
+	@rm -rf .pytest_cache
+	@find . -name '*.egg' -print0|xargs -0 rm -rf --
+	@rm -rf .eggs/
+	@rm -fr build/
+	@rm -fr dist/
+	@rm -fr *.egg-info
+
+.PHONY: help
+help: # Display target comments in 'make help'
```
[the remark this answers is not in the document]
fschuch: Solved in [6e41dd8](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/6e41dd86cf215188a947c25d1e5c579597d8f079)

jobbergate-docs/src/core.rst
matheushent: > Have you built it locally to review? Thanks for this level of attention to detail.

I did. Looks great btw. I liked how sphinx used the docstrings to generate the documentation. Do you know if there are ways to create symlinks in the docstrings that sphinx will identify?

fschuch: It is possible, see it produced a link for this cross reference: https://github.com/omnivector-solutions/jobbergate/blob/6e41dd86cf215188a947c25d1e5c579597d8f079/jobbergate-core/jobbergate_core/auth/core.py#L33

jobbergate-core/jobbergate_core/auth/core.py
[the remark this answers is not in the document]
matheushent: yes, I understand. I also thought in something like:

`
These values depend on the identity provider used for authentication. Consult your system administrator or contact the support for any guidance.
`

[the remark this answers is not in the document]
matheushent: it took my attention because seeing the documentation built, this text kind of jumps out
image

jobbergate-core/jobbergate_core/auth/core.py
fschuch: solved in [9698a75](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/9698a75a2f5cfe2f7e8c24136486f3e5fb4e5bea)

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+
+    Note:
+        This class can interoperate with the tokens generated by the ``jobbergate-cli`` package, as long
+        as they are stored in the same cache directory.
+
+    Examples:
+
+        The following example shows how to use the JobbergateAuth class to authenticate a request:
```
matheushent, line 58:
```suggestion
        The following example shows how to use the :meth:`JobbergateAuth` class to authenticate a request:
```

jobbergate-core/jobbergate_core/auth/core.py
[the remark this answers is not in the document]
fschuch: > yes, I understand. I also thought in something like:
> 
> `These values depend on the identity provider used for authentication. Consult your system administrator or contact the support for any guidance.`

That is better, thanks. Let's see if we get more comments on this topic.

jobbergate-core/jobbergate_core/auth/core.py
fschuch: solved in [281a81e](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/281a81ee6bffe72be79079ffa3676f75f87485a9)

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def acquire_tokens(self):
+        """
+        High-level method to acquire a valid access token.
+
+        This method will attempt, in order:
+
+        * Load the tokens from the cache directory (see :meth:`JobbergateAuth.load_from_cache`)
+        * If the access token is unavailable or expired, refresh both tokens
+          using the refresh token (see :meth:`JobbergateAuth.refresh_tokens`)
```
matheushent, line 111:
```suggestion
          using the refresh token grant type (see :meth:`JobbergateAuth.refresh_tokens`)
```

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+        * If the refresh token is unavailable or expired, login to fetch both tokens
```
matheushent, line 112: `Fetch` isn't the most appropriate word IMO.
```suggestion
        * If the refresh token is unavailable or expired, login to generate both tokens
```

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+        """
+        logger.debug("Authenticating request")
+
+        self.acquire_tokens()
+        access_token = self.check_credentials(token_type=TokenType.ACCESS)
+
+        request.headers["Authorization"] = f"Bearer {access_token.content}"
+        return request
+
+    def acquire_tokens(self):
```
matheushent, line 103: I personally prefer to make explicit, but I understand the preference in case you don't want to change.
```suggestion
    def acquire_tokens(self) -> None:
```

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+            token_type (TokenType): The type of token to check (default is ``access``).
+
+        Returns:
+            Token: The token is returned for reference.
+
+        Raises:
+            AuthenticationError: If the credentials are invalid.
+        """
+        token = cast(Token, self.tokens.get(token_type))
+        AuthenticationError.require_condition(token, f"{token_type} token was not found")
```
matheushent, line 139: technically saying, I see some mismatch between the docstring and the line 139. You're checking just if the `token` variable is equivalent to the literal `True`, however, the docstring implies that the function will check the token availability.

Token availability, as far as I understand, means the access token is valid. Do you think the same? If so, then the function should check if the token has the proper content that makes it available as an access token.

For example, the token `whatever` would pass the assertion but it isn't available for usage.

A proper verification here, IMO, is to use armasec to verify if the token can be authenticated against the input domain.

Thoughts?

fschuch: By availability, I mean the given `token_type` is on the internal dictionary that stores them. Since the dictionary starts empty, without `load_from_cache` and/or `login` they are not available.

The current implementation relies on `dict.get()` returning `None` for a key that is not there, it is a bit confusing indeed. I suggest the following change to clarify this:

```
        AuthenticationError.require_condition(token_type in token, f"{token_type} token was not found")
        token = self.tokens[token_type]
```

Regarding the check for tokens' expiration, some applications relying on them may run for a long time, I mean more than their lifespan, so I think it is good to check it at runtime when they are needed. In this way, automatic refresh can take place, for instance.

The point about checking if they are valid using armasec is also good. I see situations like the users changing environments (prod/staging/QA) when the tokens will be available and will be inside their lifespan, but won't be able to access the resources. How difficult do you think it would be? Could it be addressed in a new task instead of in this PR?

matheushent: most likely in other PR I think, specially because the `TokenSecurity` class from Armasec depends on the FastAPI request, which is not the case here. Possibly we would need to copy some code from there instead of pluging Armasec directly

fschuch: Code readability improved in [7f05745](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/7f05745d84b9f1e70e95ecdb1d3bb1dafc0586eb)

fschuch: More details were added on [e7ebf7b](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/e7ebf7b7e4ce8687c48b89c759ea255e98e7e58c)

jobbergate-docs/src/core.rst
fschuch: More details were added on [e7ebf7b](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/e7ebf7b7e4ce8687c48b89c759ea255e98e7e58c)

jobbergate-core/README.rst
```diff
@@ -0,0 +1,15 @@
...
+=================
+
+Jobbergate-core is a sub-project that contains the key components and logic that is shared among all other sub-projects (CLI, API, and Agent). Additionally, jobbergate-core exists to support custom automation built on top of Jobbergate.
+
+License
+-------
+* `MIT <LICENSE>`_
+
+
+Copyright
```
dusktreader, line 12: We should not call out the copyright her since we are already referencing the license which contains this information. Perhaps a link to CHANGELOG would be more appropriate.

jobbergate-core/jobbergate_core/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""
+Jobbergate-core contains key components that are shared among sub-projects.
+"""
+from .auth import AuthenticationError, JobbergateAuth, TokenError
```
dusktreader, line 4: I think we should not use relative paths in projects. I would be much more comfortable with these paths appearing as:
```
from jobbergate_core.auth import ...
```

Using the full path does not much and it removes any possible ambiguity.

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
+"""
```
dusktreader, line 1: We shouldn't have a module named "core" that is a part of a package named "jobbergate-core". It's too confusing.

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+@dataclass
+class JobbergateAuth:
+    """
+    High-level class used to handle authentication in Jobbergate.
+
+    After an instance of this class is created, it can be used to authenticate requests
+    from both `requests`_  and `httpx`_ packages by passing it to the ``auth``
+    parameter on the request (see examples bellow).
```
dusktreader, line 30:
```suggestion
    parameter on the request (see examples below).
```

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+
+    It works just out of the box. Behind the scenes, this procedure calls the
```
dusktreader, line 32:
```suggestion
    It just works out of the box. Behind the scenes, this procedure calls the
```

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+    will try to refresh them if they are expired, or will provide an URL to the user to login on the system.
+
+    Notice all steps above are also available individually as public methods, allowing a fine control for
+    advanced users.
+
+    .. _requests: https://requests.readthedocs.io/en/latest/
+    .. _httpx: https://www.python-httpx.org/
+
+    Arguments:
+        cache_directory (Path): Directory to be used for the caching tokens.
```
dusktreader: These aren't arguments, though, they are class attributes.

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+        login_domain (str): Domain used for the login.
+        login_audience (str): Audience of the login.
+        login_client_id (str): Client ID used for login.
+        tokens (Dict[TokenType, Token]): Dictionary holding the tokens needed for authentication.
```
dusktreader: I don't love the type hints in the docstring since they are already defined in the class. This seems like it could be a synchronization problem if the type of one of these attributes changes but updating the docs is forgotten.

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+        )
+        >>> response.raise_for_status()
+        >>> print(f"response = {response.json()}")
+    """
+
+    cache_directory: Path
+    login_domain: str
+    login_audience: str
+    login_client_id: str
+    tokens: Dict[TokenType, Token] = field(default_factory=dict)
```
dusktreader, line 84: Since there are only two token types, why not have each be its own attribute?
```python
access_token: Token
refresh_token: Token
```

I don't think using a dict here is helpful.

jobbergate-core/jobbergate_core/auth/core.py
```diff
@@ -0,0 +1,316 @@
...
+        """
+        logger.debug("Preparing to login to Jobbergate")
+        login_info = self._get_login_information()
+        response = self._wait_for_login_confirmation(login_info)
+        self._process_tokens_from_response(response)
+        logger.success("Login completed")
+
+    def _wait_for_login_confirmation(self, login_info: _LoginInformation) -> httpx.Response:
+        print(f"Login Here: {login_info.verification_url}")
+        while True:
```
dusktreader, line 214: I think this needs to be a limited loop. If it fails to get login after some finite amount of time, it should just fail.

jobbergate-core/tests/auth/conftest.py
```diff
@@ -0,0 +1,35 @@
+import pendulum
+import pytest
+from jose.jwt import encode
+
+
+@pytest.fixture(scope="session")
+def time_now():
```
dusktreader, line 7: You may be interested in the plummet library for testing with pendulum: https://github.com/dusktreader/plummet. The `frozen_time()` context manager allows you to freeze time at any given date.

jobbergate-core/pyproject.toml
```diff
@@ -0,0 +1,79 @@
+[tool.poetry]
+name = "jobbergate-core"
+version = "3.4.3"
+description = "Jobbergate Core"
+authors = ["Omnivector Solutions <info@omnivector.solutions>"]
+license = "MIT"
+readme = "README.rst"
+repository = "https://github.com/omnivector-solutions/jobbergate"
+packages = [ { include = "jobbergate_core" } ]
```
dusktreader, line 9: Including a `packages` section is only necessary if the package name isn't the same only with dashes replaced by underscores. In this case it wouldn't be needed.

dusktreader (changes requested):
Really nice work. I have some suggestions, questions, and comments for you.

jobbergate-core/jobbergate_core/auth/core.py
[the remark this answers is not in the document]
fschuch: The note was updated with your suggestion in [700ad32](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/700ad32409f79be7a6483f32fa959954de32cd57)

jobbergate-core/README.rst
fschuch: It is just following the patter of the project. since it is included in the README for CLI and API. Should I remove all of them?

jobbergate-core/pyproject.toml
fschuch: solved in [8328c22](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/8328c2296dee7b226a46536345949df3752182db)

jobbergate-core/jobbergate_core/auth/core.py
fschuch: Good point, it was renamed to handler and the class inside it to JobbergateAuthHandler ([8328c22](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/8328c2296dee7b226a46536345949df3752182db)).

jobbergate-core/jobbergate_core/auth/core.py
fschuch: solved in [7d94777](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/7d9477776ec7fce9deea844be884ddcd06bef054)

jobbergate-core/jobbergate_core/__init__.py
fschuch: solved in [7d94777](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/7d9477776ec7fce9deea844be884ddcd06bef054)

jobbergate-core/jobbergate_core/auth/core.py
fschuch: [705826b](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/705826b5482c89b424aa285a0cd38a4cc6f428e9)

jobbergate-core/jobbergate_core/auth/core.py
fschuch: [705826b](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/705826b5482c89b424aa285a0cd38a4cc6f428e9)

jobbergate-core/jobbergate_core/auth/core.py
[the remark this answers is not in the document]
matheushent: seems good to me, thank you

jobbergate-core/jobbergate_core/auth/core.py
matheushent: See the image below:
image

Visually saying, it is very good IMO to have the type hints in the docstring. The following example doesn't have the type hints:
image

For me, the way Felipe did is far better because it helps people with any sort of visual disability. Specially when there are many parameters, trying to find the type hint in the second example is terrible IMO. See the following for example:
image

jobbergate-core/jobbergate_core/auth/core.py
matheushent: I prefer a TypedDict here

jobbergate-core/jobbergate_core/auth/core.py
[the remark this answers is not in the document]
jamesbeedy: Small correction,`These values depend on the identity provider used for authentication. Consult your system administrator or contact Omnivector support <support@omnivector.solutions> for further guidance. `

jobbergate-core/jobbergate_core/auth/core.py
fschuch: solved in [de353c7](https://github.com/omnivector-solutions/jobbergate/pull/228/commits/de353c7382c3a765f9954042b2776d5fa716744c).

Do you feel this is enough?

jobbergate-core/jobbergate_core/auth/core.py
fschuch: In terms of the number of code lines, they are not very different, I mean:

```python
        for token in self.tokens.values():
            token.save_to_cache()
```

takes the same two as:

```
        access_token.save_to_cache()
        refresh_token.save_to_cache()
```

One potential problem is that they can be nullable in the present implementation. For instance, just the access token is necessary to authenticate a request, and just the refresh token is necessary to refresh both of them. In addition, the Handler starts without any of them, `load_from_cache` of some other method is required to fill them in.

I think this is easier to handle in the dictionary.

Notice there is no need for end-users to access the tokens directly, this choice should not impact them. The Handler takes care of everything.

jobbergate-core/jobbergate_core/auth/core.py
fschuch: I agree with both. On internal methods, I agree this is just duplication. On the other hand, I see value here where we are exposing them on the documentation, as Tosta pointed out.

Can we agree on removing them on internal details and keeping them on what is on the documentation? @dusktreader  @matheushent

matheushent: > Can we agree on removing them on internal details and keeping them on what is on the documentation?
I didn't understand, sorry. Did you mean to keep as is?

jobbergate-core/jobbergate_core/auth/core.py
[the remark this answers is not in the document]
fschuch: Done

jobbergate-core/jobbergate_core/auth/core.py
fschuch: I see no problem in removing it from initialization arguments and making it private `_tokens`.

jobbergate-core/jobbergate_core/auth/core.py
matheushent: agreed

jobbergate-core/jobbergate_core/auth/core.py
fschuch: They are arguments for the `__init__` method that is handled by `dataclass`. Also, they look better than attributes on the auto-produced documentation

jobbergate-core/README.rst
dusktreader: No, let's leave it like this for now. Something to think about for the future.

jobbergate-core/jobbergate_core/auth/core.py
dusktreader: There is still a synchronization problem. Any time you have to remember to update documentation when you change code is an opportunity to have docs go out of date. For example you could leave the functionality of a function the same such that you think you don't need to change the documentation about what the function does. However, if you change the type of an argument, you have to remember to update the docstring as well. This sort of 'drift' between implementation and documentation can be a big problem.

I will resolve this comment for now. However, I would like to see if we can find a tool that knows how to incorporate type-hints into parameter docs without requiring the user to annotate it in the docstring.

jobbergate-core/jobbergate_core/auth/core.py
dusktreader: > In terms of the number of code lines, they are not very different, I mean:
> 
> ```python
>         for token in self.tokens.values():
>             token.save_to_cache()
> ```
> 
> takes the same two as:
> 
> ```
>         access_token.save_to_cache()
>         refresh_token.save_to_cache()
> ```


I understand what you are saying. However, you have the added complexity of a loop and a data structure (dict) when you only have two items and you know there won't be any additional items later. I think the overhead of using a data structure here for two elements is unnecessary and I would even call it a code smell.

jobbergate-core/tests/auth/conftest.py
dusktreader: The only point I will make is that it's good to be able to see the timestamp where time is being frozen in the test code. This is especially true if you are comparing a `now()` value to some timestamp in test data. Otherwise, you're asking editors and reviewers to look up the project-wide freeze time in another test source file.

jobbergate-core/jobbergate_core/auth/core.py
fschuch: Ok, working on it

dusktreader (approved):

PR #232, opened 2023-03-27
PFT-1080 Made fks nullable to allow deletes

dusktreader:
#### What
Makes the foreign keys in applications and job-scripts nullable so related columns can be deleted.

#### Why
To allow deletes to happen without forcing the user to delete dependent entities.

`Task`: https://app.clickup.com/t/18022949/PFT-1080

---

#### Peer Review


fschuch (approved):

jobbergate-api/jobbergate_api/apps/applications/routers.py
```diff
@@ -162,6 +163,16 @@ async def application_delete(
     """
     Delete application from the database and S3 given it's id.
```
matheushent, line 164: quick fix in the docstring while we're here :)
```suggestion
    Delete application from the database and S3 given its id.
```

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
```diff
@@ -216,6 +217,15 @@ async def job_script_delete(job_script_id: int = Query(..., description="id of t
     """
     Delete job_script given its id.
     """
+    logger.debug(f"Unlinking job_submissions with links to job_script {job_script_id=}")
```
matheushent, line 220: same comment I made previously applies here

jobbergate-api/jobbergate_api/tests/apps/applications/test_routers.py
```diff
@@ -262,23 +264,32 @@ async def test_delete_application__fk_error(
...
-    ):
-        response = await client.delete(f"/jobbergate/applications/{inserted_id}")
-    assert response.status_code == status.HTTP_409_CONFLICT
-    error_data = json.loads(response.text)["detail"]
-    assert error_data["message"] == "Delete failed due to foreign-key constraint"
-    assert error_data["table"] == "job_scripts"
-    assert error_data["pk_id"] == f"{inserted_id}"
+    )
+    count = await database.fetch_all("SELECT COUNT(*) FROM job_scripts")
+    assert count[0][0] == 1
```
matheushent, line 275: Assert if there's only one job script in the database doesn't guarantee the job script is the correct one. There should be assertions against each row of the job script IMO

dusktreader: I agree with you. However, this mechanism is used throughout the Jobbergate API code base. If we change it in one place, we should change it throughout. However, that's some overhead I don't want to take on right now. So, I just followed the already established pattern in the new tests.

jobbergate-api/jobbergate_api/apps/job_scripts/routers.py
dusktreader: Resolved in commit a308098

jobbergate-api/jobbergate_api/apps/applications/routers.py
```diff
@@ -162,6 +163,16 @@ async def application_delete(
...
     """
+    logger.debug(f"Unlinking job_scripts with links to application {application_id=}")
```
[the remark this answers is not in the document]
dusktreader: Resolved in commit a308098

PR #273, opened 2023-06-29
PENG-1404 Add multi-tenancy to Jobbergate

dusktreader:
Adds multi-tenancy support to Jobbergate


jobbergate-api/pyproject.toml
```diff
@@ -28,21 +28,21 @@ python-multipart = "^0.0.5"
...
-asyncpg = "^0.22.0"
+asyncpg = "^0.27"
 loguru = "^0.6.0"
 uvicorn = "^0.15.0"
 armasec = "^0.11.0"
 sentry-sdk = "^1.5.0"
 yarl = "^1.7.2"
 alembic = "^1.7.5"
 typer = "^0.4.0"
-databases = {extras = ["postgresql"], version = "^0.5.5"}
```
fschuch, line 39: I know we've decided to replace databases with sqlalchemy on Jobbergate 4.x, but for the multi-tenancy feature, was databases a particular problem?

jobbergate-api/pyproject.toml
```diff
@@ -72,6 +71,9 @@ psycopg2 = "^2.9.5"
 [tool.poetry.scripts]
 dev-tools = "dev_tools:app"
 
+[tool.poetry.group.dev.dependencies]
```
fschuch, line 74: Another group for dev dependecies? I see `[tool.poetry.dev-dependencies]` above.

jobbergate-api/jobbergate_api/storage.py
```diff
@@ -3,56 +3,177 @@
...
+        If the database url is already in the engine map, return the engine stored there. Otherwise, build
+        a new one, store it, and return the new engine.
+        """
+        db_url = build_db_url(
+            override_db_name=override_db_name,
+            force_test=settings.DEPLOY_ENV.lower() == "test",
+        )
+        if db_url not in self.engine_map:
+            self.engine_map[db_url] = create_async_engine(db_url, pool_pre_ping=True)
+        return self.engine_map[db_url]
```
fschuch, line 97: I think we can avoid building the URL every time if we store the engine using `override_db_name`
```suggestion
        if db_url not in self.engine_map:
            db_url = build_db_url(
                override_db_name=override_db_name,
                 orce_test=settings.DEPLOY_ENV.lower() == "test",
            )
            self.engine_map[override_db_name] = create_async_engine(db_url, pool_pre_ping=True)
        return self.engine_map[override_db_name]
```

What do you think?

fschuch: Another approach could be

```suggestion
        return self.engine_map.setdefault(
            override_db_name,
            create_async_engine(
                build_db_url(
                    override_db_name=override_db_name,
                    force_test=settings.DEPLOY_ENV.lower() == "test",
                ),
                pool_pre_ping=True,
            ),
        )
```

fschuch (changes requested):
Well done, I left a few comments

jobbergate-api/jobbergate_api/storage.py
```diff
@@ -3,56 +3,177 @@
...
+
+    engine_map: Dict[str, AsyncEngine]
+
+    def __init__(self):
+        """
+        Initialize the EngineFactory.
+        """
+        self.engine_map = dict()
+
+    async def cleanup(self):
```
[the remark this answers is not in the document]
dusktreader: Sure thing.

jobbergate-api/pyproject.toml
dusktreader: Given the way we have to create engines on demand and make an async session for each request, it seemed like it would be easier to just make the jump. Since we couldn't use the global `database` module any more, there really wan't any utility left for the `databases` package.

jobbergate-api/pyproject.toml
dusktreader: Oh, I didn't notice that happened! I will correct it.

jobbergate-api/jobbergate_api/storage.py
dusktreader: Well, it's trickier because the `build_db_url` method pulls from settings when the `override_db_name` is None. I didn't really like one of the entries in the map potentially being keyed by `None`. Additionally, I found it handy for introspection to see the full URL for each engine.

As for the `setdefault` approach, I also considered this. However, the arguments aren't lazily evaluated, so the `create_async_engine` method would be called every single time even if there was already an engine for that key.

jobbergate-api/jobbergate_api/storage.py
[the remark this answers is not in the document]
dusktreader: Addressed in commit ff3f979

jobbergate-api/pyproject.toml
dusktreader: Addressed in commit ff3f979

jobbergate-api/pyproject.toml
fschuch: Yeah, thanks for that, the refactor to sqlalchemy looks great

jobbergate-api/jobbergate_api/storage.py
fschuch: Good point, I see no problem in keeping the code as it is

fschuch (approved):

matheushent (approved):

fschuch (approved):

fschuch (approved):

fschuch (approved):

jamesbeedy:
🔥 🔥 🔥 🔥

PR #293, opened 2023-08-01
ASP-3608 Update slurp to the new data model

fschuch:
#### What
* The data migration script is updated to work with the new data model
* The migration script is tested on local dev setup

#### Why
As a Jobbergate maintainer, I need to update the data migration script to the new data model on Jobbergate to proceed with the migration from version 3.x to 4.x.

`Task`: https://jira.scania.com/browse/ASP-3608

---

#### Peer Review


jobbergate-api/alembic/versions/20230801_071440--3bc095975b53_initial_migration_for_jobbergate_4.py
```diff
@@ -45,55 +45,59 @@ def upgrade():
     )
     op.create_table(
         "job_script_template_files",
-        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("parent_id", sa.Integer(), nullable=False),
```
fschuch, line 48: Side note: Is there a way to verify this on the API's test suite? All tests were passing, but the migration file was outdated.

dusktreader: The unit tests don't use the migrations. They use `create_all()` to create the empty tables. Perhaps this is a good opportunity for an integration test?

dusktreader (approved):

jobbergate-api/alembic/versions/20230801_071440--3bc095975b53_initial_migration_for_jobbergate_4.py
fschuch: I'll keep an eye on it, but this shouldn't be a problem for this PR, so I'm resolving the comment.

PR #305, opened 2023-08-11
PENG-1546 Guarantee only one entrypoint file per job script

fschuch:
#### What
Add business logic to guarantee job scripts have only one entry point

#### Why
Entrypoint is the file sent to slurmrest, so more than one is ambiguous and should not be allowed.

`Task`: https://app.clickup.com/t/18022949/PENG-1546

---

#### Peer Review

### DIFF
--- a/jobbergate-api/jobbergate_api/apps/job_scripts/services.py
+++ b/jobbergate-api/jobbergate_api/apps/job_scripts/services.py
@@ -135,15 +135,14 @@ class JobScriptFileService(FileService):
             raise_exc_class=ServiceError,
             raise_kwargs=dict(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY),
         )
-        await self.validate_entrypoint_file(parent_id, filename, file_type)
+        if file_type == FileType.ENTRYPOINT:
+            await self.validate_entrypoint_file(parent_id, filename)
         return await super().upsert(parent_id, filename, upload_content, **upsert_kwargs)
 
-    async def validate_entrypoint_file(self, parent_id: int, filename: str, file_type: FileType | str):
+    async def validate_entrypoint_file(self, parent_id: int, filename: str):
         """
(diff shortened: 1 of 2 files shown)


jobbergate-api/jobbergate_api/apps/job_scripts/services.py
```diff
@@ -49,6 +52,51 @@ class JobScriptFileService(FileService):
...
+        )
+        await self.validate_entrypoint_file(parent_id, filename, file_type)
+        return await super().upsert(parent_id, filename, upload_content, **upsert_kwargs)
+
+    async def validate_entrypoint_file(self, parent_id: int, filename: str, file_type: FileType | str):
+        """
+        Validate that the entrypoint file is unique.
+        """
+        if file_type != FileType.ENTRYPOINT:
+            return
```
dusktreader, line 79: It think it would make more sense to put this in the `upsert()` method so that we don't validate at all if it's not an entrypoint file.

dusktreader (changes requested):
One comment for you.

jobbergate-api/jobbergate_api/apps/job_scripts/services.py
fschuch: Do you mean just the if? Or additional logic?

dusktreader: Yeah. I mean, if we know outside this function if it's not an entrypoint file, there's no need to call the function.

fschuch: good point, solved in [be1c4a1](https://github.com/omnivector-solutions/jobbergate/pull/305/commits/be1c4a15434f7c75615d9c5316d01a15cd046c5f)

dusktreader (approved):

PR #306, opened 2023-08-14
Automatically clean up unused job scripts

fschuch:
#### What
* An automated process is created to perform the cleanup
* Job scripts that have not been used in one week are archived
* Job scripts that have been archived for one month are deleted

#### Why
As a jobbergate users, I want my unused job scripts to automatically be cleaned up if they are not used for some time so that stale job scripts don't clutter the system.

`Task`: https://jira.scania.com/browse/ASP-3417

---

#### Peer Review

### DIFF
--- a/jobbergate-api/jobbergate_api/apps/job_scripts/services.py
+++ b/jobbergate-api/jobbergate_api/apps/job_scripts/services.py
@@ -62,7 +62,10 @@ class JobScriptCrudService(CrudService):
         result = AutoCleanResponse(archived=set(), deleted=set())
 
         subquery = (
-            select(JobSubmission.job_script_id, func.max(JobSubmission.created_at).label("last_used"))  # type: ignore
+            select(
+                JobSubmission.job_script_id,
+                func.max(JobSubmission.created_at).label("last_used"),
+            )  # type: ignore
             .group_by(JobSubmission.job_script_id)
             .subquery()
         )
--- a/jobbergate-api/jobbergate_api/apps/job_scripts/services.py
+++ b/jobbergate-api/jobbergate_api/apps/job_scripts/services.py
@@ -1,12 +1,24 @@
 """Services for the job_scripts resource, including module specific business logic."""
-from typing import Any
+from typing import Any, NamedTuple
 
-from sqlalchemy import update
+import pendulum
+from loguru import logger
+from sqlalchemy import delete, func, select, update
 
 from jobbergate_api.apps.job_scripts.models import JobScript, JobScriptFile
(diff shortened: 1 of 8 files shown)


jobbergate-api/jobbergate_api/apps/job_scripts/services.py
```diff
@@ -62,7 +62,10 @@ async def auto_clean_unused_job_scripts(self) -> AutoCleanResponse:
         result = AutoCleanResponse(archived=set(), deleted=set())
 
         subquery = (
-            select(JobSubmission.job_script_id, func.max(JobSubmission.created_at).label("last_used"))  # type: ignore
+            select(
+                JobSubmission.job_script_id,
+                func.max(JobSubmission.created_at).label("last_used"),
+            )  # type: ignore
```
dusktreader, line 68: Should this function set a limit since it's sorting by last used? That way, it only grabs N most recent jobs?

jobbergate-api/tests/apps/test_services.py
```diff
@@ -74,16 +74,14 @@ async def test_create__success(
...
+
+        instance = await dummy_crud_service.create(
+            name="test-name",
+            description="test-description",
+            owner_email=tester_email,
+        )
 
         assert instance.owner_email == tester_email
-        assert instance.created_at in window
-        assert instance.updated_at in window
```
dusktreader, line 86: This is best for most test cases. For some reason, I get a very occasional off-by-one error for the window test. I have tried to isolate it, but failed. I try to limit the window tests to only methods where it's important to verify that the timestamps are being used correctly.

dusktreader (approved):
One small suggestion and a comment for you. Feel free to ignore the suggestion if it's not applicable. Nice work!

jobbergate-api/jobbergate_api/apps/job_scripts/services.py
fschuch: Is it sorting? The goal here is to get the newest job-submission for each job-script, so we can archive/delete it

jobbergate-api/tests/apps/test_services.py
fschuch: Yeah, it used to fail randomly, and it broke when I changed the time control to `pendulum` to facilitate the tests. Since it was just used in this test, I removed it. We can always go back and refactor it to work with pendulum, at least it would be more precise.

jobbergate-api/jobbergate_api/apps/job_scripts/services.py
dusktreader: Oh, I misread it. It's selecting the max. disregard.

PR #320, opened 2023-08-24
Fix: make name optional on update job-script

fschuch:
#### What
Make name optional when updating a job-script

#### Why
Bug fix.

---

#### Peer Review

### DIFF
--- a/jobbergate-cli/jobbergate_cli/subapps/job_scripts/app.py
+++ b/jobbergate-cli/jobbergate_cli/subapps/job_scripts/app.py
@@ -224,7 +224,7 @@ def update(
         ...,
         help="The id of the job script to update",
     ),
-    name: str = typer.Option(
+    name: Optional[str] = typer.Option(
         None,
         help="Optional new name of the job script.",
     ),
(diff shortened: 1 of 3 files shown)


jobbergate-cli/jobbergate_cli/subapps/job_scripts/app.py
```diff
@@ -224,7 +224,7 @@ def update(
         ...,
         help="The id of the job script to update",
     ),
-    name: str = typer.Option(
+    name: Optional[str] = typer.Option(
```
dusktreader, line 227: Use modern type hinting here:
```suggestion
    name: str | None = typer.Option(
```

dusktreader (changes requested):

dusktreader (approved):

jobbergate-cli/jobbergate_cli/subapps/job_scripts/app.py
fschuch: It is valid if we add `from __future__ import annotations`, but it still results in some internal errors in `typer`, so reverting the suggestion.

PR #321, opened 2023-08-25
Patch s3 bucket on multitenancy

fschuch:
#### What
Overwrite the bucket name on S3 when multitenancy is enabled.

#### Why
Fix bug.

---

#### Peer Review


jobbergate-api/jobbergate_api/apps/services.py
```diff
@@ -40,15 +41,20 @@ def __init__(self, message, status_code=status.HTTP_400_BAD_REQUEST, **kwargs):
 class DatabaseBoundService:
     """
     Provide base class for services that bind to a database session.
+
+    This class holds a reference to the session and provides methods to bind and unbind the session.
+    It also keeps track of all instances of the service so that they can be iterated over.
     """
 
     _session: AsyncSession | None
+    _database_instances: WeakSet = WeakSet()
```
dusktreader, line 50: I don't think this needs to be a "private" attribute. Also, you don't need the `iter_instances` method because WeakSet is iterable. Also, the name is misleading. Why not just name it "services"?

dusktreader (changes requested):

jobbergate-api/jobbergate_api/apps/services.py
fschuch: Good points. I'll make the changes. One point with the name is that FileService inherited from both, so we'd better avoid naming conflicts: `FileService(DatabaseBoundService, BucketBoundService`

fschuch: Addressed in [3fbdeef](https://github.com/omnivector-solutions/jobbergate/pull/321/commits/3fbdeef971bdf4c4e89a01a0cf6f6a63cab7629d)

Notice I agree a name like `services` would be cleaner, but it leads to conflict, so a prefix was added to avoid that.

dusktreader (approved):
Very nice

PR #367, opened 2023-10-04
Potential fix to an issue with git dependencies

dusktreader:
Fixed some issues with git dependencies and local packages for docker-compose.


jobbergate-agent/pyproject.toml
```diff
@@ -21,7 +21,7 @@ packages = [{ include = "jobbergate_agent" }]
 
 [tool.poetry.dependencies]
 python = "^3.8"
-jobbergate-core = { path = "../jobbergate-core", develop = true }
+jobbergate-core = "4.0"
```
fschuch, line 24: I'm concerned this relaxes the dependency and may lead to the agent and core being installed out of sync on the clusters, and that may lead to scenarios that were not covered on tests. The current implementation pins them to the exact same version number at building time, resulting in a more reliable approach. 

It means that if we install jobbergate-agent 4.5.6 from PyPI, jobbergate-core 4.5.6 and just 4.5.6 will be installed as a dependency, not 4.x.

I see the agent is installed on the container using `poetry install`, which includes the dev dependency:
`jobbergate-core = { path = "../jobbergate-core", develop = true }`

So, I feel like the line above would make both docker and our deployment automation happy.

fschuch (changes requested):
Good work on that.

jobbergate-agent/pyproject.toml
dusktreader: This is a typo. It should be:
```
jobbergate-core = "^4.0"
```
That will make sure we get any patches to core throughout 4.0.X lifecycle.

For regular installs (not with dev group) we shouldn't use relative paths. Instead, it should use a fixed version. When we publish the agent to pypi and install it from there, it should not include with it source code from jobbergate-core. If there are updates to core that need to be published, they should be published under a new version and the dependency from the agent should be bumped to the new version.

What we really don't want to happen is to have a version of the agent published that includes changes to core that haven't been packaged as a part of a release that has been tested and fully vetted.

fschuch (approved):

PR #399, opened 2023-10-26
[ASP-4070] Distinguish application-id, job-script-id and job-submission-id

julianaklulo:
#### What
Add alias to `application-id`, `job-script-id` and `job-submission-id` to distinguish each other when submitting a job.

#### Why
All of them were displayed as `id`, which was confusing the users.

`Task`: https://jira.scania.com/browse/ASP-4070

---

#### Peer Review

### DIFF
--- a/jobbergate-cli/jobbergate_cli/schemas.py
+++ b/jobbergate-cli/jobbergate_cli/schemas.py
@@ -127,7 +127,7 @@ class ApplicationResponse(pydantic.BaseModel, extra=pydantic.Extra.ignore):
     Describes the format of data for applications retrieved from the Jobbergate API endpoints.
     """
 
-    id: int
+    application_id: int = pydantic.Field(alias="id")
     name: str
     owner_email: str
     created_at: datetime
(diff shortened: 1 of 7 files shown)


matheushent (commented):
what's the point of distinguishing this way? IMO it just adds redundancy.

## EDIT
I understand the confusion for users. Disregard my comment

jobbergate-cli/jobbergate_cli/schemas.py
```diff
@@ -127,7 +127,7 @@ class ApplicationResponse(pydantic.BaseModel, extra=pydantic.Extra.ignore):
     Describes the format of data for applications retrieved from the Jobbergate API endpoints.
     """
 
-    id: int
+    application_id: int = pydantic.Field(alias="id")
```
fschuch, line 130: I wonder if we could apply an alias when exporting the model, so the code can stay as it is, but the names presented to the users are different.

Just thinking aloud, because the implementation on this PR is good to go as well.

fschuch (changes requested):
Nice work, I left a few comments

fschuch (approved):

PR #411, opened 2023-11-09
[ASP-4229] Add info to docs about terminal window

julianaklulo:
#### What
Include section suggesting to use the terminal maximized to avoid issues with messages being cropped out.

#### Why
The `inquirer` library used in the questions has a limitation of single line messages, which can cause issues in terminals without enough width.

`Task`: https://jira.scania.com/browse/ASP-4229

---

#### Peer Review

### DIFF
--- a/jobbergate-docs/docs/source/elements/apps/cli.md
+++ b/jobbergate-docs/docs/source/elements/apps/cli.md
@@ -137,3 +137,8 @@ provide the following sub-commands:
 Details for each subcommand can be viewed by passing the `--help` flag to any of them.
 
 Use the `--help` option to explore the CLI and disccover the usage and options for all the subcommands.
+
+
+## Usability
+
+To ensure that you can see the full output of the CLI, we recommend that you use a terminal in a maximized window.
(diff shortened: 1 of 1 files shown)


jobbergate-docs/docs/source/elements/apps/cli.md
```diff
@@ -137,3 +137,8 @@ provide the following sub-commands:
 Details for each subcommand can be viewed by passing the `--help` flag to any of them.
 
 Use the `--help` option to explore the CLI and disccover the usage and options for all the subcommands.
+
+
+## Usability
```
fschuch, line 142: I think this could be a note to get a little more attention from the users: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#+type:note

Also, what do you think of citing it is a constraint by a dependency and adding that link for reference?

fschuch (changes requested):

jobbergate-docs/docs/source/elements/apps/cli.md
julianaklulo: Does this look better?
image

fschuch: Yes, thanks

fschuch (approved):

PR #417, opened 2023-11-15
PENG-1910 Fixed recurring failures in CI pipeline

dusktreader:
I moved the session management code in the storage module into its own context manager. Then, I modified the conftest to override this particular part of the code only. Now, the storage code is minimally aware of special test logic (it only blows up if the session context manager is used in a test) and the particulars of managing a session for a test are isolated in the conftest.py

I also eliminated the async specific tests. Because the test fixtures force routes to use a preset session, you can't call multiple routes simultaneously due to this sqlalchemy error: https://docs.sqlalchemy.org/en/20/errors.html#error-isce. Any validation of concurrency should occur as part of an integration test instead.

### DIFF
--- a/jobbergate-api/CHANGELOG.md
+++ b/jobbergate-api/CHANGELOG.md
@@ -3,6 +3,8 @@
 This file keeps track of all notable changes to jobbergate-api
 
 ## Unreleased
+- Moved database session management into a dedicated context manager and removed test-aware logic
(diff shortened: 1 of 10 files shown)


julianaklulo (approved):

fschuch (approved):
Good job. I know it is internal logic, but the sessions are a key part on the API's logic, so please consider adding a new entry on the changelog so we can keep track if it.

PR #419, opened 2023-11-23
[ASP-4245] Update agent download logic file

julianaklulo:
#### What
Add the env var `DOWNLOAD_JOB_SCRIPT` to control whether the agent should download the job script file.
If it's set to `True`, the behaviour is exactly as before.
If it's set to `False`, the submission will be rejected if there are supporting files. If there aren't, the job script will be passed to the `slurmrestd` API as the payload and won't be downloaded to the submit dir.

#### Why
The submission was failing when the agent didn't have permission to write to the submit dir.

`Task`: https://jira.scania.com/browse/ASP-4245

---

#### Peer Review

### DIFF
--- a/jobbergate-agent/jobbergate_agent/jobbergate/submit.py
+++ b/jobbergate-agent/jobbergate_agent/jobbergate/submit.py
@@ -48,23 +48,89 @@ def get_job_parameters(slurm_parameters: Dict[str, Any], **kwargs) -> SlurmJobPa
     return SlurmJobParams.parse_obj(merged_parameters)
 
 
-async def retrieve_job_script_file(job_script_file: JobScriptFile, submit_dir: Path):
+async def retrieve_submission_file(file: JobScriptFile) -> str:
     """
-    Get a submission file from the backend.
-
-    Write the file to the submit_dir if DOWNLOAD_JOB_SCRIPTS is set to True.
-    Return the decoded file content.
+    Get a submission file from the backend and return the decoded file content.
     """
-    response = await backend_client.get(job_script_file.path)
+    response = await backend_client.get(file.path)
     response.raise_for_status()
 
-    if SETTINGS.DOWNLOAD_JOB_SCRIPTS:
-        local_script_path = submit_dir / job_script_file.filename
-        local_script_path.parent.mkdir(parents=True, exist_ok=True)
-        local_script_path.write_bytes(response.content)
-        logger.debug(f"Copied file to {local_script_path}")
-
     return response.content.decode("utf-8")
+
+
+async def write_submission_file(file_content: str, filename: str, submit_dir: Path):
+    """
+    Write a decoded file content to the submit_dir.
+    """
+    local_script_path = submit_dir / filename
+    local_script_path.parent.mkdir(parents=True, exist_ok=True)
+    local_script_path.write_bytes(file_content.encode("utf-8"))
+
+    logger.debug(f"Copied file to {local_script_path}")
+
+
+async def process_supporting_files(pending_job_submission: PendingJobSubmission, submit_dir: Path):
+    """
+    Process the submission support files.
+
+    Write the support files to the submit_dir if WRITE_SUBMISSION_FILES is set to True.
+    Reject the submission if there are support files with WRITE_SUBMISSION_FILES set to False.
+    """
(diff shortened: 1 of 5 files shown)


matheushent (approved):
LGTM

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
```diff
@@ -45,6 +48,25 @@ def get_job_parameters(slurm_parameters: Dict[str, Any], **kwargs) -> SlurmJobPa
     return SlurmJobParams.parse_obj(merged_parameters)
 
 
+async def retrieve_job_script_file(job_script_file: JobScriptFile, submit_dir: Path):
```
fschuch, line 51: Nice touch. I suggest annotating the returning type as well.

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
```diff
@@ -72,26 +94,40 @@ async def submit_job_script(
 
         submit_dir = pending_job_submission.execution_directory or SETTINGS.DEFAULT_SLURM_WORK_DIR
 
-        job_script = None
-
-        for metadata in pending_job_submission.job_script.files:
-            local_script_path = submit_dir / metadata.filename
-            local_script_path.parent.mkdir(parents=True, exist_ok=True)
+        """
```
fschuch, line 97: This comment block may be an indication that the file's logic could be encapsulated in a function, what do you think?

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
```diff
@@ -72,26 +94,40 @@ async def submit_job_script(
...
 
         JobSubmissionError.require_condition(
             job_script,
             "Could not find an executable script in retrieved job script data.",
         )
 
+        if SETTINGS.DOWNLOAD_JOB_SCRIPTS:
+            # Download the supporting files to the execution dir
+            files_to_download = [retrieve_job_script_file(file, submit_dir) for file in supporting_files]
+            await asyncio.gather(*files_to_download)
```
fschuch, line 121: This is also nice, to speed up the submissions. I feel job_scripts could be in this gather as well, to also take advantage of asyncio. Some changes may be required in the design here, but I hope it will simplify a bit the code.

jobbergate-agent/jobbergate_agent/settings.py
```diff
@@ -50,6 +50,9 @@ class Settings(BaseSettings):
     TASK_JOBS_INTERVAL_SECONDS: int = Field(60, ge=10, le=3600)  # seconds
     TASK_GARBAGE_COLLECTION_HOUR: Optional[int] = Field(None, ge=0, le=23)  # hour of day
 
+    # Job submission settings
+    DOWNLOAD_JOB_SCRIPTS: bool = True
```
fschuch, line 54: Maybe it should refer to saving the files on disk? For clarification, because they are downloaded from the backend anyway to send to slurm restd.

fschuch (changes requested):

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
julianaklulo: Done!

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
julianaklulo: The main problem is that the supporting files need to be written but we don't care about their decoded data, as for the entrypoint we need the decoded data to pass as the payload.
I changed the logic to encapsulate everything related to the job script entrypoint in its own function.

jobbergate-agent/jobbergate_agent/settings.py
julianaklulo: Renamed it to `WRITE_SUBMISSION_FILES`.

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
fschuch: Good work on that

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
fschuch: That is a fair point. It is looking good after the encapsulation.

fschuch (approved):

PR #430, opened 2023-12-04
[ASP-3966] Implement pagination

julianaklulo:
#### What
Implement pagination for `list_all` commands for each subapp: `job-script-templates`, `job-scripts` and `job-submissions`.

#### Why
The users couldn't see all the results because only the first page of results was being rendered.

`Task`: https://jira.scania.com/browse/ASP-3966

---

#### Peer Review


jobbergate-cli/jobbergate_cli/render.py
```diff
@@ -202,3 +202,43 @@ def render_single_result(
...
+            style_mapper = StyleMapper()
+        for key in first_row.keys():
+            table.add_column(key, **style_mapper.map_style(key))
+
+        for row in filtered_results:
+            table.add_row(*[str(v) for v in row.values()])
+
+        console = Console()
+        console.print()
+        console.print(table)
```
fschuch, line 244: If you agree, the indentation can be reduced here:

```suggestion
        return
    if envelope.total == 0:
        terminal_message("There are no results to display", subject="Nothing here...")
        return

    current_page = envelope.page
    total_pages = envelope.pages

    if ctx.full_output or hidden_fields is None:
        filtered_results = envelope.items
    else:
        filtered_results = [{k: v for (k, v) in d.items() if k not in hidden_fields} for d in envelope.items]

    first_row = filtered_results[0]

    table = Table(title=title, caption=f"Page {current_page} of {total_pages} - Total items: {envelope.total}")
    if style_mapper is None:
        style_mapper = StyleMapper()
    for key in first_row.keys():
        table.add_column(key, **style_mapper.map_style(key))

    for row in filtered_results:
        table.add_row(*[str(v) for v in row.values()])

    console = Console()
    console.print()
    console.print(table)
```

jobbergate-cli/jobbergate_cli/render.py
```diff
@@ -202,3 +202,43 @@ def render_single_result(
...
+        total_pages = envelope.pages
+
+        if ctx.full_output or hidden_fields is None:
+            filtered_results = envelope.items
+        else:
+            filtered_results = [{k: v for (k, v) in d.items() if k not in hidden_fields} for d in envelope.items]
+
+        first_row = filtered_results[0]
+
+        table = Table(title=title, caption=f"Page {current_page} of {total_pages} - Total items: {envelope.total}")
```
fschuch, line 233: Could we include the number of entries per page?

jobbergate-cli/jobbergate_cli/subapps/pagination.py
```diff
@@ -0,0 +1,95 @@
...
+    assert jg_ctx.client is not None
+
+    current_page = 1
+
+    while True:
+        envelope = cast(
+            ListResponseEnvelope,
+            make_request(
+                jg_ctx.client,
+                url_path + f"?page={current_page}",
```
fschuch, line 30: I think that `page` could be added on `params` to make the code more readable, do you agree?

jobbergate-cli/poetry.lock
```diff
@@ -1,9 +1,10 @@
-# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.4.0 and should not be changed by hand.
```
fschuch, line 1: I see no changes to `pyproject.toml`, why do we have changes on `poetry.lock`?

fschuch (changes requested):
Nice work. I left a few comments for you.

jobbergate-cli/jobbergate_cli/render.py
julianaklulo: I modified it to make it easier to read.

jobbergate-cli/poetry.lock
julianaklulo: The previous iteration of the prototype was using some external dependencies (pynput, for instance), but it was removed for the final version.

jobbergate-cli/jobbergate_cli/render.py
julianaklulo: Yep, I added it.

jobbergate-cli/jobbergate_cli/subapps/pagination.py
```diff
@@ -0,0 +1,99 @@
...
+                    inquirer.List(
+                        "navigation",
+                        message=message,
+                        choices=choices,
+                        default=PaginationChoices.NEXT_PAGE,
+                    )
+                ]
+            )
+
+        if answer["navigation"] == PaginationChoices.NEXT_PAGE:
```
fschuch, line 94: It breaks if the user exits with `CTRL+C`, it looks like `answer=None` in this case. Could you handle this to avoid a big traceback on the screen?

fschuch:
Could we change the three listing commands to sort descending by `id` by default? Now with pagination, the unsorted by default is bugging me 😅. I think is makes more sense to show more recent first.

fschuch (approved):

PR #459, opened 2024-01-26
PENG-2045 Fixed cluster_name presentation in multi-tenancy mode

dusktreader:
* Now, the user will see "cluster_name" when they view job_submissions instead of "client_id"
* When run in "multi-tenancy" mode, the organization_id is removed from the cluster_name when rendered

### DIFF
--- a/jobbergate-cli/jobbergate_cli/auth.py
+++ b/jobbergate-cli/jobbergate_cli/auth.py
@@ -94,6 +94,29 @@ def validate_token_and_extract_identity(token_set: TokenSet) -> IdentityData:
             log_message=f"Token data could not be extracted to identity: {err}",
             original_error=err,
         )
+
+    if settings.MULTI_TENANCY_ENABLED:
+        try:
+            org_dict = token_data["organization"]
+            JobbergateCliError.require_condition(
+                len(org_dict) == 1,
+                message="Did not find exactly one organization in token payload",
+            )
+            identity.organization_id = list(org_dict.keys())[0]
+        except Exception as err:
+            raise Abort(
+                """
+                The access token is invalid.
+
+                Please try logging in again.
+                """,
+                subject="Invalid access token",
+                support=True,
+                log_message="Organization payload required in multi-tenancy mode is malformed ",
+                sentry_context=dict(token_data=token_data),
+                original_error=err,
+            )
+
     return identity
 
 
--- a/jobbergate-cli/jobbergate_cli/requests.py
+++ b/jobbergate-cli/jobbergate_cli/requests.py
@@ -250,7 +250,7 @@ def make_request(
     if response_model_cls is None:
         return data
 
-    logger.debug("Validating response data with ResponseModel")
+    logger.debug(f"Validating response data with {response_model_cls}")
     try:
         return response_model_cls(**data)
     except pydantic.ValidationError as err:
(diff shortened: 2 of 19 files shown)


jobbergate-cli/jobbergate_cli/auth.py
```diff
@@ -94,6 +94,29 @@ def validate_token_and_extract_identity(token_set: TokenSet) -> IdentityData:
...
+        try:
+            org_dict = token_data["organization"]
+            JobbergateCliError.require_condition(
+                len(org_dict) == 1,
+                message="Did not find exactly one organization in token payload",
+            )
+            identity.organization_id = list(org_dict.keys())[0]
+        except Exception as err:
+            raise Abort(
+                """
```
fschuch, line 108: The error message could be a bit more specific, so we can quickly identify the a possible problem when providing support to end users

jobbergate-cli/jobbergate_cli/requests.py
```diff
@@ -250,7 +250,7 @@ def make_request(
     if response_model_cls is None:
         return data
 
-    logger.debug("Validating response data with ResponseModel")
+    logger.debug(f"Validating response data with {response_model_cls}")
```
fschuch, line 253: Good catch!

jobbergate-cli/poetry.lock
```diff
@@ -1,9 +1,10 @@
-# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.4.1 and should not be changed by hand.
```
fschuch, line 1: I see no change in dependencies on pyproject.toml. Could you revert the changes in this file?

jobbergate-cli/jobbergate_cli/subapps/pagination.py
```diff
@@ -14,8 +17,10 @@ def handle_pagination(
     abort_message: str = "There was an error communicating with the API",
     params: Optional[Dict[str, Any]] = None,
     title: str = "Results List",
-    style_mapper: StyleMapper = None,
-    hidden_fields: Optional[List[str]] = None,
+    style_mapper: StyleMapper | None = None,
+    hidden_fields: List[str] | None = None,
+    nested_response_model_cls: Type[pydantic.BaseModel] | None = None,
```
fschuch, line 22: Nice touch. That was indeed missing.

fschuch (changes requested):
Noce work. I left a few comments.

jobbergate-cli/jobbergate_cli/auth.py
dusktreader: We don't want to give the user too much information so that it's confusing or troubling. If the command is run in verbose mode, you will see the log_message below displayed. Additionally, Sentry should catch this error and provide more context for us as well.

jobbergate-cli/poetry.lock
dusktreader: Rebased, removed, and force-pushed in commit a5b7946.

jobbergate-cli/jobbergate_cli/auth.py
fschuch: Fair enough

fschuch (approved):

PR #460, opened 2024-01-26
[ASP-3335] Add clone capability for Applications and Job Scripts

fschuch:
#### What
- Added column cloned_from_id to three base tables
- Added clone_instance to generic services
- Added clone_instance to file services
- Added endpoint to clone job templates
- Added endpoint to clone job scripts
- Added command to clone a job template
- Added command to clone a job script

#### Why
As a Jobbergate user, I would like to have the ability to clone an existing Application or Job Script so that I can adapt it for my own use without modifying the original.

`Task`: https://jira.scania.com/browse/ASP-3335

---

#### Peer Review


jobbergate-api/jobbergate_api/apps/models.py
```diff
@@ -58,10 +60,22 @@ class IdMixin:
 
     Attributes:
         id: The id of the job script template.
+        cloned_from: Specify the id of the row that this row was cloned from.
     """
 
     id: Mapped[int] = mapped_column(Integer, primary_key=True)
 
+    @declared_attr
+    def cloned_from_id(cls) -> Mapped[int | None]:
```
fschuch, line 69: This needs to be added to the detailed and maybe list views on the response schemas, right now there is no way to visualize this value.

dusktreader, line 69: Why is this specified as a `declared_attr`?

jobbergate-cli/jobbergate_cli/subapps/applications/app.py
```diff
@@ -405,3 +405,89 @@ def download_files(
...
+    ctx: typer.Context,
+    id: Optional[int] = typer.Option(
+        None,
+        help=f"The specific id of the application. {ID_NOTE}",
+    ),
+    identifier: Optional[str] = typer.Option(
+        None,
+        help=f"The human-friendly identifier of the application. {IDENTIFIER_NOTE}",
+    ),
+    application_identifier: Optional[str] = typer.Option(
```
dusktreader, line 422: I wonder if these params should be like `new_application_identifier` and such to make it more clear that these will override the old attributes upon cloning.

jobbergate-cli/jobbergate_cli/subapps/applications/app.py
```diff
@@ -405,3 +405,89 @@ def download_files(
...
+        None,
+        help="Optional new application identifier to override the original",
```
dusktreader, line 424: I think this text should also explain that the new identifier cannot match the old one.

dusktreader (changes requested):

jobbergate-api/jobbergate_api/apps/models.py
fschuch: solved in [885ab05](https://github.com/omnivector-solutions/jobbergate/pull/460/commits/885ab0582416b667b3f2f69e48febeb59547e47c)

fschuch: [SQLAlchemy requires](https://docs.sqlalchemy.org/en/20/orm/declarative_mixins.html) it in a mixing class so that we can refer to another column from it. Otherwise, the error message is very specific:

```text
sqlalchemy.exc.InvalidRequestError:
Columns with foreign keys to other columns must be declared as @declared_attr callables on declarative mixin classes.
```

jobbergate-cli/jobbergate_cli/subapps/applications/app.py
fschuch: I considered it, but I chose this path to be consistent with the parameter names we use on `update-application`, and to require less typing from the users. If you agree, I think we'd better keep them in this way.

jobbergate-cli/jobbergate_cli/subapps/applications/app.py
fschuch: Solved in [110b589](https://github.com/omnivector-solutions/jobbergate/pull/460/commits/110b58961ea8c1694ef26da8476ef5ad6ace3784).

jobbergate-api/jobbergate_api/apps/models.py
dusktreader: Oh, gotcha. Nice.

dusktreader (approved):

PR #469, opened 2024-02-02
PENG-2064 Capture job state info in jobbergate

dusktreader:
The fundamental feature of this change is to capture the slurm job state in the API so that it can be observed by users. The major change here is a remapping of the job_submissions statuses so that they no longer overlap with job_states reported by slurm and instead are only used to manage how jobbergate components interact with job submissions along their lifecycle. New columns are added for the job_info and the job_state. These capture the details of a job's state as it is processed.

Change Summary
 - Jobbergate API
   - Added columns for slurm_job_state and slurm_job_info
   - Added mappings for slurm_job_state details (long descriptions and abbreviations)
   - Updated schemas for job payloads
   - Created specialized submit and reject endpoints so that agent doesn't need to understand job_submission statuses
   - Added a migration for the database changes including existing status mapping
   - Updated and added unit tests
 - Jobbergate Agent
   - Renamed the finish.py module to update.py
   - Changed logic such that all active jobs have their job state updated on each pass
   - Moved all logic from the api.py module to update.py and submit.py
   - Updated schemas for revised job payloads
   - Updated and added unit tests
 - Jobbergate CLI changes
   - Updated schemas for revised job payloads
   - Made slurm_job_info hidden by default on job_submissions
 - Jobbergate Docs changes
   - Slight docs updates for the agent app
   - Slight docs updates for the tutorial


examples/simple-job-script.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+#SBATCH -J dummy_job
+#SBATCH -t 60
+
+echo "Executing simple job script"
+
+echo "Simple output from $HOSTNAME" > /nfs/simple-output.txt
```
fschuch, line 8: I'm not a big fan of using the absolute path in this case because it ignores the submission directory, which can lead to file conflicts. 
```suggestion

 echo "Simple output from $HOSTNAME" > ./simple-output.txt
```

jobbergate-agent/jobbergate_agent/jobbergate/schemas.py
```diff
@@ -94,17 +94,12 @@ class SlurmSubmitResponse(pydantic.BaseModel, extra=pydantic.Extra.ignore):
     job_id: Optional[int]
 
 
-class SlurmSubmittedJobStatus(pydantic.BaseModel, extra=pydantic.Extra.ignore):
+class SlurmJobData(pydantic.BaseModel, extra=pydantic.Extra.ignore):
```
fschuch, line 97: Could we remove the prefixes `slurm_job` from the fields? The class already makes clear they detail a slurm job, I don't think we need repetition on the fields

jobbergate-agent/jobbergate_agent/jobbergate/update.py
```diff
@@ -0,0 +1,95 @@
...
+    logger.debug(f"Fetching slurm job status for slurm job {slurm_job_id}")
+
+    response = await slurmrestd_client.get(f"/job/{slurm_job_id}")
+
+    with SlurmrestdError.handle_errors("Failed to fetch job state from slurm", do_except=log_error):
+        response.raise_for_status()
+        data = response.json()
+
+    jobs = data["jobs"]
+    SlurmrestdError.require_condition(
```
fschuch, line 25: I think this would be a good case to mark the jobs as `unknow` status. Because if slurm doesn't know about them and we keep the status as it was, they will be continues pulled by the agent just to raise this error again.

jobbergate-api/alembic/versions/20240130_105238--ec2d2948fb41_remap_job_submissions_statuses.py
```diff
@@ -0,0 +1,201 @@
...
+            "COMPLETED",
+            "FAILED",
+            "UNKNOWN",
+            "CANCELLED",
+            name="jobsubmissionstatus",
+        ),
+        type_=sa.Enum(
+            "CREATED",
+            "SUBMITTED",
+            "REJECTED",
```
fschuch, line 120: Do we need three of them? From the agent's perspective, they are all kind of done, it does not need to worry about them anymore. The new columns with slurm status will hold the details differentiating them.

fschuch (changes requested):
Hey, goof job. I left a few comments for you.

jobbergate-agent/jobbergate_agent/jobbergate/schemas.py
dusktreader: Well, the reason that we are capturing it in this way is because the API requires the prefixes on its endpoints. For the API, this makes sense since it is storing them in the job_submissions table. There, having the `slurm_job` prefix makes sense.

jobbergate-api/alembic/versions/20240130_105238--ec2d2948fb41_remap_job_submissions_statuses.py
dusktreader: I think it's useful to have a distinction in the status for which jobs completed with a normal exit code and those that were not completed for some reason. Also, REJECTED means that the job was never processed by Slurm at all but was rejected at submission time. The statuses have distinct meanings.

examples/simple-job-script.sh
dusktreader: Updated in commit e2e8554.

jobbergate-agent/jobbergate_agent/jobbergate/schemas.py
dusktreader: Partly resolved in commit e2e8554. The fields retain the prefix on the API side, but in the agent code, the prefixes are removed.

jobbergate-agent/jobbergate_agent/jobbergate/update.py
dusktreader: Updated in commit e2e8554.

jobbergate-api/alembic/versions/20240130_105238--ec2d2948fb41_remap_job_submissions_statuses.py
```diff
@@ -0,0 +1,201 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "job_submissions",
+        sa.Column(
+            "slurm_job_state",
+            sa.Enum(
```
fschuch, line 25: It looks like `UNKNOWN` is missing here

fschuch (approved):

jobbergate-api/alembic/versions/20240130_105238--ec2d2948fb41_remap_job_submissions_statuses.py
dusktreader: Added it in latest commit

fschuch (approved):

PR #472, opened 2024-02-05
[ASP-3984] Add status badges for test coverage on Jobbergate

fschuch:
#### What
* Setup Jobbergate to export test coverage reports to [codecov](https://about.codecov.io).
* Add new status badge to the repo: [codecov](https://codecov.io/gh/omnivector-solutions/jobbergate)

#### Why
As a Jobbergate maintainer, I want to add a test coverage status badge to Jobbergate's repository. This will allow stakeholders, project contributors, and end-users to easily see the project's test coverage.

`Task`: https://jira.scania.com/browse/ASP-3984

#### References

* Using flags to manage mono-repos in codecov: https://docs.codecov.com/docs/flags
* Easily upload coverage reports to Codecov from GitHub Actions: https://github.com/codecov/codecov-action

---

#### Peer Review

### DIFF
--- a/jobbergate-agent/pyproject.toml
+++ b/jobbergate-agent/pyproject.toml
@@ -62,7 +62,7 @@ freezegun = "^1.2.2"
 isort = "^5.9.3"
 mypy = "^0.931"
 pytest-asyncio = "^0.18.2"
-pytest-cov = "^3.0.0"
+pytest-cov = "^4.0.0"
 pytest-env = "^0.6.2"
 pytest-mock = "^3.7.0"
 pytest-random-order = "^1.0.4"
@@ -75,8 +75,7 @@ asyncio_mode = "auto"
 addopts = [
     "--random-order",
     "--cov=jobbergate_agent",
-    "--cov-report=term-missing",
-    "--cov-fail-under=90",
+    "--cov-report=xml:tests/coverage.xml",
 ]
 env = [
     "JOBBERGATE_AGENT_OIDC_DOMAIN = auth.com",
@@ -84,6 +83,10 @@ env = [
     "JOBBERGATE_AGENT_OIDC_CLIENT_SECRET = DUMMY-TEST-CLIENT-SECRET",
     "JOBBERGATE_AGENT_SLURMRESTD_JWT_KEY_STRING = DUMMY-JWT-SECRET",
 ]
+
+[tool.coverage.report]
+fail_under = 90
+show_missing = true
 
 [tool.black]
 line-length = 120
--- a/jobbergate-api/pyproject.toml
+++ b/jobbergate-api/pyproject.toml
@@ -50,7 +50,6 @@ auto-name-enum = "^2.0.0"
 [tool.poetry.group.dev.dependencies]
 asgi-lifespan = "^1.0.1"
 black = "^23"
-coverage = { extras = ["toml"], version = "^7.2.7" }
 flake8-docstrings = "^1.6.0"
 ipython = "^7.31.1"
 isort = "^5.9.3"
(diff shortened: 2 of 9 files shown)


jobbergate-composed/docker-compose.yml
```diff
@@ -57,6 +57,7 @@ services:
       - ../jobbergate-api/jobbergate_api/:/app/jobbergate_api/
       - ../jobbergate-api/dev_tools/:/app/dev_tools/
       - ../jobbergate-api/tests/:/app/tests/
+      - ../jobbergate-api/:/app/
```
fschuch, line 60: We need to take the file `coverage.xml` out of the container. @dusktreader, do you agree with this as an alternative to solve the issue?

matheushent: not asked to me but... I agree. Indeed, I do that in other projects across the company

jobbergate-api/pyproject.toml
```diff
@@ -92,8 +92,14 @@ ignore = "W503,D200,D106,D402"
...
 
 [tool.pytest.ini_options]
-minversion = "6.0"
-addopts = ["--random-order", "--cov=jobbergate_api"]
+minversion = "7.0"
+addopts = [
+    "--random-order",
+    "--cov=jobbergate_api",
+    "--cov-report=term-missing",
+    "--cov-fail-under=90",
```
dusktreader, line 100: Why use command line arguments here instead of using the configuration blocks you removed below?

jobbergate-composed/docker-compose.yml
dusktreader: I'm not a big fan of volume mounting the whole project folder into the container. When the running app creates files in this folder, they are owned by root and can cause pain. If we are going to do this kind of mount, there's no point in mounting the source code, test folder, and dev_tools separately.

dusktreader (commented):
Left some questions for you.

jobbergate-agent/pyproject.toml
```diff
@@ -76,6 +76,7 @@ addopts = [
     "--cov=jobbergate_agent",
     "--cov-report=term-missing",
     "--cov-fail-under=90",
+    "--cov-report=xml",
```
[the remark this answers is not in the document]
fschuch: Codecov centrally ingests `.xml`, `.json` and `.txt` type coverage report formats, so it looked appropriate.

jobbergate-api/pyproject.toml
fschuch: Just for consistency across the components. The three others define them using command-line arguments, so I changed the API to do the same.

jobbergate-composed/docker-compose.yml
fschuch: That is a good point. Do you have a suggestion to copy just `coverage.xml` back from the container?

dusktreader: You could change the path so that it is in a different folder that you can more safely volume mount.

jobbergate-api/pyproject.toml
dusktreader: I think it would actually be better to go the other way and start using configuration as opposed to command line arguments for the other sub-projects as well.

jobbergate-composed/docker-compose.yml
fschuch: I redirected the report to `tests/coverage.xml`, so there is no need for mounting the whole project. I used the same path on other components for consistency.

Addressed in [0efb40c](https://github.com/omnivector-solutions/jobbergate/pull/472/commits/0efb40c7cab3e5e71b4ad8429a7c22915a182ca9).

jobbergate-api/pyproject.toml
fschuch: Following your suggestions, I changed the API and other components to use configuration blocks.

As a side note, I found no way to set the report format on the blocks. If I try `format="xml"` it complains when loading the config that the value should be boolean. It is probably a bug in the package since `format=true` results in an error when writing the report file because the format `true` is not supported.

We will stick to `--cov-report=xml:tests/coverage.xml` for now.

dusktreader (approved):

PR #479, opened 2024-02-13
PENG-2039 Added support for notifications via rabbitmq

dusktreader:
Added logic to send a notfication on a rabbitmq channel when a job terminates with a status of:

- REJECTED
- ABORTED
- COMPLETED

Also added unit tests.
Also added rabbitmq to jobbergate-composed for help with testing.


fschuch (changes requested):
Nice work, I left a few comments

jobbergate-api/jobbergate_api/rabbitmq_notification.py
```diff
@@ -0,0 +1,87 @@
...
+
+    logger.debug("Publishing status change to notification queue")
+
+    try:
+        message_payload = dict(
+            path=f"jobs.job_submissions.{job_submission.id}",
+            user_email=job_submission.owner_email,
+            action="status",
+            additional_context=dict(
+                status=job_submission.status,
```
matheushent, line 72: I believe the following is more informative. Also, for some purposes of ours, it is good to have the job submission id in the payload as well (even known it is in the path).

```suggestion
                job_status=job_submission.status,
                job_submission_id=....,
```

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
```diff
@@ -259,15 +260,22 @@ async def job_submission_agent_update(
...
-        # The job is completed and should notify the user...when we have notifications for Jobbergate
         update_dict["status"] = JobSubmissionStatus.DONE
 
     job_submission = await secure_services.crud.job_submission.update(id, **update_dict)
 
+    if job_submission.status in (
+        JobSubmissionStatus.ABORTED,
+        JobSubmissionStatus.DONE,
+    ):
+        await publish_status_change(
```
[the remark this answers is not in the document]
matheushent: > Could it be done by a background tasks? The response could be sent quicker, there is no need to wait for the notification to be processed.

the notification itself takes milliseconds to execute, so I don't see a solid reason to make it in a background task. Also, the possible error is in case the RabbitMQ endpoint is not available, and for that, there are other methods to avoid (such as define dependencies on docker compose or liveness probe in a Kubernetes environment)

jobbergate-api/jobbergate_api/rabbitmq_notification.py
```diff
@@ -0,0 +1,87 @@
...
+@asynccontextmanager
+async def rabbitmq_connect(
+    exchange_name=None,
+    do_purge=False,
+):
+    if exchange_name is None:
+        exchange_name = settings.RABBITMQ_DEFAULT_EXCHANGE
+
+    event_loop = asyncio.get_running_loop()
+    connection = await aio_pika.connect_robust(
```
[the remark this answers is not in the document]
matheushent: I think is fine. I do that in other micro services of ours and the performance is pretty wondrous.

jobbergate-api/jobbergate_api/rabbitmq_notification.py
```diff
@@ -0,0 +1,87 @@
...
+                message=aio_pika.Message(
+                    body=json.dumps(message_payload).encode("utf-8"),
+                    delivery_mode=aio_pika.DeliveryMode.PERSISTENT,
+                    headers=dict(organization=organization_id),
+                ),
+                routing_key="status",
+            )
+    except Exception as exc:
+        logger.error("Failed to publish status change notification")
+        logger.exception(exc)
```
matheushent, line 87: doesn't mypy complain about missing a return statement here?

jobbergate-api/CHANGELOG.md
```diff
@@ -3,9 +3,10 @@
 This file keeps track of all notable changes to jobbergate-api
 
 ## Unreleased
+- Added notifications via rabbitmq for job status updates [PENG-2039]
 
 
-## 4.3.0a7 -- 2024-02-08
+## 4.3.0a7 -- 2024-02-0
```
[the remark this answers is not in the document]
dusktreader: Looks like I accidently deleted a character. Resolved in commit ee30c67.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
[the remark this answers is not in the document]
dusktreader: So, first, it doesn't need to be a background task because, like Tosta mentioned, it is very, very fast to publish a message to Rabbitmq and it is all async.

Second, all dangerous logic in the `publish_status_chanbe` method is wrapped by a try/catch that logs an error and continues. So, if the publish fails for some reason, the response will not be impacted.

jobbergate-api/jobbergate_api/rabbitmq_notification.py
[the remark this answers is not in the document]
dusktreader: My observation aligns with Tosta's. It's performant.

jobbergate-api/jobbergate_api/rabbitmq_notification.py
dusktreader: The field is called `status` in all contexts. If we changed it to `job_status` it would be too easy to confuse with the slurm job state. This is the status of the `job_submission`, not the job itself (that is provided in `slurm_job_state`).

I'm not sure how it would help to have the job submission id in the payload as well.

jobbergate-api/jobbergate_api/rabbitmq_notification.py
dusktreader: no, it does not.

jobbergate-api/jobbergate_api/rabbitmq_notification.py
matheushent: > I'm not sure how it would help to have the job submission id in the payload as well.

We have some private micro services that would benefit by this.

jobbergate-api/jobbergate_api/apps/job_submissions/routers.py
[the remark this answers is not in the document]
fschuch: Fair enough

jobbergate-api/jobbergate_api/rabbitmq_notification.py
fschuch: I'm ok with either way, so I'm approving the PR, but you need to get to an agreement on this point

fschuch (approved):

jobbergate-api/jobbergate_api/rabbitmq_notification.py
matheushent: we can leave this discussion for later.

PR #482, opened 2024-02-15
[ASP-3326] Add option to render templates locally

julianaklulo:
#### What
Add a new command in the CLI to render the job script from a local application.

#### Why
To make it easier to test applications without uploading to the API.

`Task`:  https://jira.scania.com/browse/ASP-3326

---

#### Peer Review


jobbergate-cli/jobbergate_cli/subapps/job_scripts/app.py
```diff
@@ -162,6 +163,65 @@ def create(
...
+        help="The path to the application directory to use as a template for the job script.",
+    ),
+    job_script_name: str = typer.Option(
+        ...,
+        help="The name of the job script to render locally.",
+    ),
+    output_path: pathlib.Path = typer.Option(
+        ...,
+        help="The path to the directory where the rendered job script should be saved.",
+    ),
```
fschuch, line 181: I wonder if we can reduce the typing to run it. I think we could make `application_path` an argument since it is the more important here, so it is positional and users do not need to type `--application_path`. Use would be like `jobbergate render-job-script-locally ../../examples/simple-application`.

We could also default both paths to the current working directory and have some default name for extra reduction on typing.

What do you think?

```suggested
@app.command()
@handle_abort
def render_locally(
    ctx: typer.Context,
    application_path: pathlib.Path = typer.Argument(
        pathlib.Path("."),
        help="The path to the application directory to use as a template for the job script.",
        dir_okay=True,
    ),
    job_script_name: str = typer.Option(
        "rendered-locally",
        help="The name of the job script to render locally.",
    ),
    output_path: pathlib.Path = typer.Option(
        pathlib.Path("."),
        help="The path to the directory where the rendered job script should be saved.",
    ),
```

jobbergate-cli/jobbergate_cli/subapps/job_scripts/tools.py
```diff
@@ -180,6 +188,146 @@ def get_template_output_name_mapping(config: JobbergateConfig, job_name: str) ->
...
+    mapped_template_files = {
+        new_filename: file
+        for file in app_data.template_files
+        if (new_filename := template_name_mapping.get(file.filename, None))
+    }
+
+    for new_filename, template_file in mapped_template_files.items():
+        file_content = render_template(template_file.path, param_dict_flat)
+
+        if template_file.file_type.upper() == "ENTRYPOINT" and sbatch_params:
```
fschuch, line 324: Why not use the enum here and on the schemas?

jobbergate-cli/jobbergate_cli/subapps/job_scripts/tools.py
```diff
@@ -180,6 +188,146 @@ def get_template_output_name_mapping(config: JobbergateConfig, job_name: str) ->
...
+            file_content = inject_sbatch_params(file_content, sbatch_params)
+
+        with open(output_path / new_filename, "w") as f:
+            f.write(file_content)
```
fschuch, line 328: A log message containing the filename would be useful here

jobbergate-cli/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 This file keeps track of all notable changes to jobbergate-cli
 
 ## Unreleased
+- Added option to render an application into a job script locally [ASP-3326]
```
fschuch, line 6:
```suggestion
- Added command to render an application into a job script locally [ASP-3326]
```

fschuch (changes requested):
Well done, this command is going to be very useful.

jobbergate-cli/jobbergate_cli/subapps/job_scripts/tools.py
julianaklulo: I based the `LocalTemplateFile` on the `TemplateFileResponse` and it's using `str` there instead of the enum, so I overlooked it.
But that's a good catch, I updated to use the `enum` here.

jobbergate-cli/jobbergate_cli/subapps/job_scripts/tools.py
julianaklulo: Totally, I added it.

jobbergate-cli/jobbergate_cli/subapps/job_scripts/tools.py
```diff
@@ -180,6 +188,148 @@ def get_template_output_name_mapping(config: JobbergateConfig, job_name: str) ->
...
+
+    application_source_code = app_data.workflow_files[0].path.read_text()
+
+    (app_config, app_module) = load_application_data(app_data, application_source_code)
+
+    supplied_params = validate_parameter_file(param_file) if param_file else dict()
+    execute_application(app_module, app_config, supplied_params, fast_mode=fast)
+    update_template_files_information(app_data, app_config)
+
+    param_dict_flat = flatten_param_dict(app_config.dict())
```
fschuch, line 311: The dynamic name is missing here, in case the application code sets it at runtime.

```suggestion
    param_dict_flat = flatten_param_dict(app_config.dict())

    if param_dict_flat.get("job_script_name"):
        # Possibly overwrite script name if set at runtime by the application
        job_script_name = param_dict_flat["job_script_name"]
        logger.debug("Job script name was set by the application at runtime: {}", job_script_name)
```

julianaklulo: Good catch, I added it.

fschuch (approved):

PR #497, opened 2024-02-26
[ASP-4584] Replace slurmrestd by sbatch on the API

fschuch:
#### What
Replace slurmrestd by sbatch on the API.

#### Why
This is the first component on the migration.

**Note:** PR will be merged into a feature branch.

`Task`: https://jira.scania.com/browse/ASP-4584

---

#### Peer Review


jobbergate-api/alembic/versions/20240226_175811--35ccf022687d_replace_slurmrestd_by_sbatch.py
```diff
@@ -0,0 +1,41 @@
...
+# revision identifiers, used by Alembic.
+revision = "35ccf022687d"
+down_revision = "ec2d2948fb41"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("job_submissions", sa.Column("sbatch_arguments", sa.String(), nullable=True))
```
matheushent, line 23: what do you think about this column being a JSONB column instead of a string? To me, it makes much more sense to parse the batch arguments instead of just putting them as it comes.

jobbergate-api/alembic/versions/20240226_175811--35ccf022687d_replace_slurmrestd_by_sbatch.py
```diff
@@ -0,0 +1,41 @@
...
+    op.drop_column("job_submissions", "template_vars")
```
matheushent, line 24: I think you should apply some logic here to move the template_vars to the sbatch params format before just dropping the column

jobbergate-api/alembic/versions/20240226_175811--35ccf022687d_replace_slurmrestd_by_sbatch.py
fschuch: Now, to your point. I've considered keeping it as a JSONB, having a parser may result in early error feedback to the users, and a few other points. But at the same time, it would impose constrains on the sbatch instances and versions the system as a whole can operate with.

One of the main goals in replacing slurmrestd by sbatch is:
> Extra functionality can be added to the sbatch command by installing plugins. For instance, we can cite the support to run containerized tasks added by pyxis. Such flexibility is not possible by using Slurm Rest API.

So to make the API more flexible and considering it just stores this data, I think it should be up to slurm on each cluster to decide it they are fine with the arguments or not (reject the job in this case). It is even more important when we consider the multi-tenancy scenario, where different clusters can operate on different settings and sbatch versions, and the API have to support them all.

jobbergate-api/alembic/versions/20240226_175811--35ccf022687d_replace_slurmrestd_by_sbatch.py
fschuch: It is possible, but not trivial. This PR deletes the parser that goes from SBATCH to execution parameters with +300 lines of code, we would need to code it in the opposite direction and keep it in this migration file. In addition, not all of them have a one-on-one relation, so we would need to cover some edge cases.

I also consider that dropping it won't affect any completed job.

dusktreader: wait, why is template_vars being dropped?

fschuch: Please, see this note: https://github.com/omnivector-solutions/jobbergate/pull/497#discussion_r1508172295

fschuch: > wait, why is template_vars being dropped?

Please, see this note: https://github.com/omnivector-solutions/jobbergate/pull/497#discussion_r1508172295

jobbergate-api/alembic/versions/20240226_175811--35ccf022687d_replace_slurmrestd_by_sbatch.py
fschuch: So, `--name=foo -c bar --flag` should be stored as the follow?
```json
{
  "name": "foo",
  "c": "bar",
  "flag": null,
}
```

dusktreader: It could be, yes. However, I think we need to consider how these params would be passed in to the API first. Would Jobbergate CLI just ask the user to pass in a string of sbatch params? It would be very clunky in the API to have the user just supply a blob of sbatch params in a single string, I think.

jobbergate-api/alembic/versions/20240226_175811--35ccf022687d_replace_slurmrestd_by_sbatch.py
dusktreader: I see. The column was incorrectly named in the first place. Now, we are dropping the column so we need to use the erronious name to do so. This will actually kill two birds with one stone.

jobbergate-api/alembic/versions/20240226_175811--35ccf022687d_replace_slurmrestd_by_sbatch.py
fschuch: We agreed that a list of strings is the choice here. It was addressed in [a16b72f](https://github.com/omnivector-solutions/jobbergate/pull/497/commits/a16b72f5a02d903d4bfc5a01210f83578bff66cc).

jobbergate-api/alembic/versions/20240226_175811--35ccf022687d_replace_slurmrestd_by_sbatch.py
fschuch: We agreed we were okay with just dropping the previous column.

dusktreader (approved):

PR #499, opened 2024-03-07
PENG-2094 Removed savepoints and added SQL Logging

dusktreader:
Removed the savepoints in the storage auto-session as they are not being utilized and can introduce unecessary overhead for API requests.

Also added configurable sql logging to allow more insight into queries that are being executed.


dusktreader:
Failing codecov report should be ignored. The auto-sessions are not used for tests and aren't testable. It's also negligible value to unit test logging. We have no logging tests at the moment, and it wouldn't be worth the effort to add it for this change.

fschuch (approved):

PR #518, opened 2024-03-25
PENG-2116 implement logic for auto updating the agent

matheushent:
#### What
This PR implements a task (quite in the same format as the current tasks) whose purpose is to auto update the agent based on the upstream version. Essentially, the logic fetches the openapi config available in the API at `/jobbergate/openapi.json` and tries to match against the current version of the agent.

In case the version is exactly the same, no action is taken, otherwise, there are the following scenarios:
1. The upstream version has the major different
  i. In this case, no action is taken because it is assumed the agent must be manually patched (updated or rolled back)
2. The upstream version has minor different
  i. In this case, the agent will update independently if it is an upgrade or an rollback
3. The upstream version has patch different
  i. Exact behaviour as [2]

#### Why
Every version change in the agent creates a logistical chaos because every cluster that depends on it gets outdated. This makes difficult to create automation on top of Jobbergate for many clusters.

`Task`: https://sharing.clickup.com/t/h/c/18022949/PENG-2116/F97V6RJPMF65TEW

---

#### Peer Review


jobbergate-agent/jobbergate_agent/tasks.py
```diff
@@ -7,13 +7,21 @@
...
 from jobbergate_agent.settings import SETTINGS
 from jobbergate_agent.utils.logging import log_error, logger_wraps
 from jobbergate_agent.utils.scheduler import BaseScheduler, Job
 
 
+def self_update_task(scheduler: BaseScheduler) -> Job:
+    """
+    Schedule a task to self update the agent every ``TASK_SELF_UPDATE_INTERVAL_SECONDS`` seconds.
+    """
+    return scheduler.add_job(self_update_agent, "interval", seconds=SETTINGS.TASK_JOBS_INTERVAL_SECONDS)
```
fschuch, line 22:
```suggestion
    return scheduler.add_job(self_update_agent, "interval", seconds=SETTINGS.TASK_SELF_UPDATE_INTERVAL_SECONDS)
```

A dedicated config would be useful for it. I see the docstring indicates one, but it is not reflected on the code here and on `settings.py`.

jobbergate-agent/jobbergate_agent/internals/update.py
```diff
@@ -0,0 +1,71 @@
...
+    upstream_version = await _fetch_upstream_version_info()
+    logger.debug(
+        f"Jobbergate Agent version info: current_version={current_version}, upstream_version={upstream_version}"
+    )
+
+    if _compare_versions(current_version, upstream_version):
+        logger.warning("The Jobbergate Agent is outdated in relation of the upstream version, an update is required.")
+
+        logger.debug("Shutting down the scheduler...")
+        scheduler.shutdown(wait=False)
```
fschuch, line 61: I fear this wont be enough to apply the changes. This may do the agent run with updated tasks, but the agent itself is still the one that was running initially. I think we may need to restart the entire service here.

jobbergate-agent/jobbergate_agent/internals/update.py
```diff
@@ -0,0 +1,71 @@
...
+        return False
+    current_major, _, _ = map(int, current_version.split("."))
+    upstream_major, _, _ = map(int, upstream_version.split("."))
+    if current_major != upstream_major:
+        return False
+    return True
+
+
+def _update_package(version: str) -> None:
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "--upgrade", f"{package_name}=={version}"])
```
fschuch, line 43: Nice touch. I though that selecting the correct environment to update could be tricky, but `sys.executable` solves the issue.

jobbergate-agent/jobbergate_agent/internals/update.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+async def self_update_agent():
```
fschuch, line 46: Some error handling is needed to avoid killing the agent if something goes sideways on the functions here.

jobbergate-agent/jobbergate_agent/internals/update.py
```diff
@@ -0,0 +1,71 @@
...
+    logger.debug("Fetching the upstream version info from jobbergate API")
+    response = await jobbergate_api_client.get("/jobbergate/openapi.json")
+    response.raise_for_status()
+
+    data = response.json()
+    upstream_version: str = data["info"]["version"]
+    return upstream_version
+
+
+def _compare_versions(current_version: str, upstream_version: str) -> bool:
```
fschuch, line 24: `need_update` could be a more representative name for this function

fschuch (changes requested):

jobbergate-agent/jobbergate_agent/internals/update.py
matheushent: good feedback, thank you!

jobbergate-agent/jobbergate_agent/internals/update.py
matheushent: can you elaborate a little more?

jobbergate-agent/jobbergate_agent/internals/update.py
matheushent: hmmmm, from my local testing that was actually enough because later on the function runs `schedule_tasks(scheduler)` which loads the new plugins

fschuch: Does it propagate changes unrelated to tasks, lets way on `settings.py` or some of the other files?

matheushent: hm, I won't give 100% certainty on the answer because that was difficult to test locally, however, I suppose so.

jobbergate-agent/jobbergate_agent/internals/update.py
matheushent: I think not having this was intentional by myself so we can capture in the logs as is

jobbergate-agent/jobbergate_agent/tasks.py
matheushent: see [7e288f1](https://github.com/omnivector-solutions/jobbergate/pull/518/commits/7e288f1dcece8e6321d4298875482e3092e04982)

jobbergate-agent/jobbergate_agent/internals/update.py
matheushent: see [6d2a394](https://github.com/omnivector-solutions/jobbergate/pull/518/commits/6d2a394811dbdbbb57703f3d72115e8791ce1899)

jobbergate-agent/jobbergate_agent/internals/update.py
```diff
@@ -0,0 +1,71 @@
...
+    In case the agent is updated, the scheduler is shutdown and restarted with the new version.
+    """
+    current_version = get_distribution(package_name).version
+    upstream_version = await _fetch_upstream_version_info()
+    logger.debug(
+        f"Jobbergate Agent version info: current_version={current_version}, upstream_version={upstream_version}"
+    )
+
+    if _need_update(current_version, upstream_version):
+        logger.warning("The Jobbergate Agent is outdated in relation of the upstream version, an update is required.")
```
dusktreader, line 58:
```suggestion
        logger.warning("The Jobbergate Agent is outdated in relation to the upstream version; an update is required.")
```

jobbergate-agent/jobbergate_agent/settings.py
```diff
@@ -42,6 +42,7 @@ class Settings(BaseSettings):
     # Task settings
     TASK_JOBS_INTERVAL_SECONDS: int = Field(60, ge=10, le=3600)  # seconds
     TASK_GARBAGE_COLLECTION_HOUR: Optional[int] = Field(None, ge=0, le=23)  # hour of day
+    TASK_SELF_UPDATE_INTERVAL_SECONDS: int = Field(60 * 60, ge=10)  # seconds
```
fschuch, line 45: I think that making it optional and `None` by default makes more sense. So by default system admins are sure the version installed by them is pinned. I see use cases for auto update as well, but this should be explicitly set.

matheushent: check [8f30d49](https://github.com/omnivector-solutions/jobbergate/pull/518/commits/8f30d49012ff3e40f595c30247c90225547dd421)

jobbergate-agent/jobbergate_agent/internals/update.py
matheushent: done in [8f30d49](https://github.com/omnivector-solutions/jobbergate/pull/518/commits/8f30d49012ff3e40f595c30247c90225547dd421)

fschuch (approved):

PR #527, opened 2024-04-10
PENG-2152 Coerced empty identifiers to None values

dusktreader:
Added validators to make sure that the value for `identifier` is coerced to `None` if the string is empty.

Also added a validator to make sure that the `name` field cannot be empty.

Added unit tests for the added logic.
---

#### Peer Review

### DIFF
--- a/jobbergate-api/CHANGELOG.md
+++ b/jobbergate-api/CHANGELOG.md
@@ -4,6 +4,8 @@ This file keeps track of all notable changes to jobbergate-api
 
 ## Unreleased
 
+- Added logic to coerce empty `identifier` on job script templates to a `None` value [PENG-2152]
+- Added logic disallow empty `name` on job script templates [PENG-2152]
 
 ## 5.0.0a1 -- 2024-04-04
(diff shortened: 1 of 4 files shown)


jobbergate-agent/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "jobbergate-agent"
-version = "5.0.0a1"
+version = "4.7.0"
```
fschuch, line 3: That is not expected to change in a feature branch, what happened here?

fschuch (changes requested):

jobbergate-agent/pyproject.toml
dusktreader: Oh, hell, that was leftover from QA'ing the auto-update stuff. Will restore.

dusktreader: Fixed the error, amended the commit, and force-pushed.

fschuch (approved):

PR #551, opened 2024-04-25
[ASP-4989, PENG-2208] Add jobbergate admin role to jobbergate-api

fschuch:
#### What
Added new permissions:
- `jobbergate:job-templates:admin`
- `jobbergate:job-scripts:admin`
- `jobbergate:job-submissions:admin`

They allow users to modify and delete entries that they do not own.

#### Why
Allow more control for key users on the entries on the database.

`Task`:
- https://jira.scania.com/browse/ASP-4989
- https://app.clickup.com/t/18022949/PENG-2208

---

#### Peer Review


matheushent (commented):
Overall comment: I think it'd be better to use the `lockdown_some` method from armasec instead of adding the if statements. It would keep the code cleaner IMO.

fschuch:
> Overall comment: I think it'd be better to use the `lockdown_some` method from armasec instead of adding the if statements. It would keep the code cleaner IMO.

I have considered adding it on lockdown. However, the admin in this case is not about having or not having access to the route, but bypassing the ownership check before updating/deleting entries once processing the request. A non-admin user who does not own the entry gets `status.HTTP_403_FORBIDDEN` in this case.

jobbergate-api/jobbergate_api/apps/job_script_templates/routers.py
```diff
@@ -154,9 +154,11 @@ async def job_script_template_update(
 ):
     """Update a job script template by id or identifier."""
     logger.info(f"Updating job script template {id_or_identifier=} with {update_request=}")
-    await secure_services.crud.template.get(
-        id_or_identifier, ensure_attributes={"owner_email": secure_services.identity_payload.email}
-    )
+    instance = await secure_services.crud.template.get(id_or_identifier)
+    if Permissions.JOB_TEMPLATES_ADMIN not in secure_services.identity_payload.permissions:
```
dusktreader, line 158: I think we actually want just one admin permission that covers all three entities. If someone is an admin of jobbergate, they should have the same rights on all three.

dusktreader (changes requested):

jobbergate-api/jobbergate_api/apps/job_script_templates/routers.py
fschuch: solved in [2576767](https://github.com/omnivector-solutions/jobbergate/pull/551/commits/2576767153a5df1a6e440b435aed62666a144b42)

dusktreader (approved):
This looks great! Nice work.

PR #560, opened 2024-05-22
[ASP-5192] Hide less used command line options

fschuch:
#### What
* Hide less used commands on the help text when backward compatibility mode is enabled
* Add the subcommands `applications`, `job-scripts`, and `job-submissions` to backward compatibility mode
* Signalize backward compatible commands are deprecated and indicate the new alternative

See the resulting help text:

image

#### Why
Users will be less confused by the amount of command options available to them.

`Task`: https://jira.scania.com/browse/ASP-5192

---

#### Peer Review

### DIFF
--- a/jobbergate-cli/CHANGELOG.md
+++ b/jobbergate-cli/CHANGELOG.md
@@ -4,6 +4,9 @@ This file keeps track of all notable changes to jobbergate-cli
 
 ## Unreleased
 
+* Hided less used commands on the cli when backward compatibility mode is enabled [ASP-5192]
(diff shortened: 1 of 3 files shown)


matheushent (commented):
I think the `replaced by` is nice to have but doesn't really indicate what the command does. I would prefer to keep the original help text and add the `replaced by` statement at the end; the `deprecated` indicative + the `replaced by` should be enough for the user to notice these commands should no longer be further used.

dusktreader (approved):

PR #569, opened 2024-06-17
PENG-2278 Upgraded Jobbergate Agent to Pydantic 2

dusktreader:
* Also upgraded Mypy version
* Also removed pydantic dependency in Jobbergate Core
* Also fixed some typing issues highlighted by Mypy

`Task`: https://app.clickup.com/t/18022949/PENG-2278

---

#### Peer Review


jobbergate-agent/jobbergate_agent/settings.py
```diff
@@ -3,12 +3,26 @@
...
+    """
+    Determine if running in test mode and return the correct path to the .env file if not.
+    """
+    _test_mode = "pytest" in sys.modules
+    env_file: str | None
+    if not _test_mode:
+        env_file = ".env"
+    else:
+        env_file = None
+    return env_file
```
fschuch, line 23: Optional suggestion:

```suggestion
     test_mode = "pytest" in sys.modules
     return None if test_mode else ".env"
```

fschuch (approved):
Good work. I left an optional suggestion. I also performed integration tests on jobbergate-composed and found no issues.

jobbergate-agent/jobbergate_agent/settings.py
dusktreader: It was left like this after I did some debugging. It's clear enough, I think, as it is.

PR #587, opened 2024-07-12
[ASP-4779] Jobbergate auto-login

fschuch:
#### What
Trigger the login workflow when required:

* Current workflow: User tries any command, like `jobbergate job-scripts create <...>`. If access credentials are expired or unavailable, jobbergate presents a message asking the user to log in and exists with a non-zero code. The user then runs `jobbergate login`. The user reruns the first command.

* Expected workflow: User tries any command. If access credentials are expired or unavailable, jobbergate starts the login workflow. Once login is completed, jobbergate processes the original command. 

**Details:**
- The required logic was already implemented on jobbergate-core
  - Core got its own request handler to enhance usability
  - Pydantic added to core to facilitate requests and response validation
  - Custom errors were added so client code can handle them properly
- Auth handler from core replaced the persona logic used on the cli
  - It enables auto-login
  - It makes credentials lazy loaded, so commands that make no request face no problems when tokens are unavailable/expired. 

#### Why
Required by a user as an enhancement of UX.

`Task`: https://jira.scania.com/browse/ASP-4779

---

#### Peer Review


jobbergate-cli/jobbergate_cli/config.py
```diff
@@ -93,22 +79,31 @@ def compute_extra_settings(self) -> Self:
         cache_dir = self.JOBBERGATE_CACHE_DIR
         cache_dir.mkdir(exist_ok=True, parents=True)
 
-        self.JOBBERGATE_APPLICATION_MODULE_PATH = cache_dir / constants.JOBBERGATE_APPLICATION_MODULE_FILE_NAME
-        self.JOBBERGATE_APPLICATION_CONFIG_PATH = cache_dir / constants.JOBBERGATE_APPLICATION_CONFIG_FILE_NAME
-
-        log_dir = cache_dir / "logs"
-        log_dir.mkdir(exist_ok=True, parents=True)
-        self.JOBBERGATE_LOG_PATH = log_dir / "jobbergate-cli.log"
+        return self
```
dusktreader, line 82: This method needs to have it's docstring update to reflect what it's actually doing now.

jobbergate-cli/jobbergate_cli/config.py
```diff
@@ -93,22 +79,31 @@ def compute_extra_settings(self) -> Self:
...
+    @computed_field
+    def JOBBERGATE_APPLICATION_MODULE_PATH(self) -> Path:
+        return self.JOBBERGATE_CACHE_DIR / constants.JOBBERGATE_APPLICATION_MODULE_FILE_NAME
+
+    @computed_field
+    def JOBBERGATE_APPLICATION_CONFIG_PATH(self) -> Path:
+        return self.JOBBERGATE_CACHE_DIR / constants.JOBBERGATE_APPLICATION_CONFIG_FILE_NAME
+
+    @computed_field
+    def JOBBERGATE_CLUSTER_LIST_PATH(self) -> Path:
```
dusktreader, line 105: I like what you did here in making these all `computed_fields`. It's a nice update.

jobbergate-cli/jobbergate_cli/exceptions.py
```diff
@@ -105,3 +107,28 @@ def wrapper(*args, **kwargs):
             raise typer.Exit(code=1)
 
     return wrapper
+
+
+def handle_authentication_error(func):
```
dusktreader, line 112: Since `handle_abort` is already being used on all the commands where this handler would need to be included, why not bake this logic directly into `handle_abort` and avoid having to use an extra decorator?

jobbergate-core/jobbergate_core/tools/requests.py
```diff
@@ -0,0 +1,182 @@
...
+
+ResponseModel = TypeVar("ResponseModel", bound=BaseModel)
+
+
+def deserialize_request_model(request_model: BaseModel, request_kwargs: dict[str, Any]):
+    """
+    Deserialize a pydantic model instance into request_kwargs for an httpx client request in place.
+    """
+    with check_expressions(
+        main_message="Request was incorrectly structured to include a request model instance.",
```
dusktreader, line 40: This message is confusing. It makes it sound like the error was that a model instance was included in the request.

dusktreader (changes requested):
Very nice work! Left a few comments for you.

jobbergate-cli/jobbergate_cli/config.py
fschuch: Thanks for noticing. All the mypy issues caused because they were marked as optional were bugging me, this make things more clear.

jobbergate-core/jobbergate_core/tools/requests.py
fschuch: Solved in [f706334](https://github.com/omnivector-solutions/jobbergate/pull/587/commits/f70633446ba994f2ada899d3e02ffa2fafc8ab6e)

jobbergate-cli/jobbergate_cli/config.py
fschuch: Good catch.

Since just `JOBBERGATE_LOG_PATH` was left, I changed it to a field validator for clarification. Solved in f706334.

jobbergate-cli/jobbergate_cli/exceptions.py
fschuch: I initially thought that making both explicit was better than making them implicit, but this duplication on all commands is not ideal. I modified the code a bit to centralize the error handling: f706334.

jobbergate-cli/jobbergate_cli/exceptions.py
```diff
@@ -132,3 +132,17 @@ def wrapper(*args, **kwargs):
             ) from err
 
     return wrapper
+
+
+def handle_errors(func):
```
dusktreader, line 137: Since this is only being used in one place, I think it makes sense to just move the logic into `safe_entrypoint()`. Adding an extra layer here doesn't add much I think.

dusktreader (changes requested):

jobbergate-cli/jobbergate_cli/exceptions.py
fschuch: `handle_errors` is actually on the tests to ensure we got it covered when testing the commands: https://github.com/omnivector-solutions/jobbergate/blob/00a99988bbe05a3b4f88e157892f499b9d231923/jobbergate-cli/tests/subapps/conftest.py#L28-L40

That is why it is not directly on `safe_entrypoint`.

dusktreader: it seems like you could also just include the two existing decorators here to avoid having to declare an extra decorator:

```
 @pytest.fixture 
 def make_test_app(dummy_context): 
     def _main_callback(ctx: Context): 
         ctx.obj = dummy_context 
  
     def _helper(command_name: str, command_function: Callable): 
         main_app = Typer() 
         main_app.callback()(_main_callback) 
         safe_command = handle_abort(handle_authentication_error(command_function))
         main_app.command(name=command_name)(safe_command) 
         return main_app 
```

fschuch: Suggestion addressed in [e50914c](https://github.com/omnivector-solutions/jobbergate/pull/587/commits/e50914ce8b07429de28e6aa70311abb22f9877fc).

dusktreader (approved):

PR #591, opened 2024-08-05
PENG-2348 implement an endpoint to get a cluster's status by the client_id

matheushent:
#### What
This PR implement an endpoint for fetching a cluster's status by passing the `client_id` parameter in the path of the request.

#### Why
Necessary to enhance the API consumers performance when querying specific clusters' statuses.

`Task`: https://sharing.clickup.com/t/h/c/18022949/PENG-2348/QWZFBKV72ZNL293

---

#### Peer Review

### DIFF
--- a/jobbergate-api/jobbergate_api/apps/clusters/routers.py
+++ b/jobbergate-api/jobbergate_api/apps/clusters/routers.py
@@ -62,3 +62,26 @@ async def get_cluster_status(
     logger.debug("Getting list of cluster statuses")
     query = select(ClusterStatus).order_by(ClusterStatus.client_id)
     return await paginate(secure_session.session, query)
+
+
+@router.get(
+    "/status/{client_id}",
+    description="Endpoint to get the status of a specific cluster.",
+    response_model=ClusterStatusView,
+)
+async def get_cluster_status_by_client_id(
+    client_id: str,
+    secure_session: SecureSession = Depends(
+        secure_session(Permissions.ADMIN, Permissions.CLUSTERS_READ, commit=False)
+    ),
+):
+    """
+    Get the status of a specific cluster.
+    """
+    logger.debug("Getting cluster status for client_id={}", client_id)
+    query = select(ClusterStatus).where(ClusterStatus.client_id == client_id)
+    result = await secure_session.session.execute(query)
+    instance = result.scalar_one_or_none()
+    if instance is None:
+        return FastAPIResponse(status_code=status.HTTP_404_NOT_FOUND)
+    return ClusterStatusView.model_validate(instance)
--- a/jobbergate-api/jobbergate_api/apps/clusters/routers.py
+++ b/jobbergate-api/jobbergate_api/apps/clusters/routers.py
@@ -84,4 +84,4 @@ async def get_cluster_status_by_client_id(
     instance = result.scalar_one_or_none()
     if instance is None:
         return FastAPIResponse(status_code=status.HTTP_404_NOT_FOUND)
-    return ClusterStatusView.model_validate(instance)
+    return instance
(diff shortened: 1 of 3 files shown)


jobbergate-api/jobbergate_api/apps/clusters/routers.py
```diff
@@ -62,3 +62,26 @@ async def get_cluster_status(
...
+    """
+    Get the status of a specific cluster.
+    """
+    logger.debug("Getting cluster status for client_id={}", client_id)
+    query = select(ClusterStatus).where(ClusterStatus.client_id == client_id)
+    result = await secure_session.session.execute(query)
+    instance = result.scalar_one_or_none()
+    if instance is None:
+        return FastAPIResponse(status_code=status.HTTP_404_NOT_FOUND)
+    return ClusterStatusView.model_validate(instance)
```
dusktreader, line 87: I don't believe it's necessary to do `model_validate()` here as FastAPI should do this automatically.

jobbergate-api/tests/apps/clusters/test_routers.py
```diff
@@ -125,3 +125,34 @@ async def test_get_cluster_status__bad_permission(self, client, inject_security_
...
+    async def test_get_cluster_status_by_client_id__not_found(
+        self, permission, client_id, client, inject_security_header, synth_session
+    ):
+        inject_security_header("who@cares.com", permission)
+
+        response = await client.get(f"/jobbergate/clusters/status/{client_id}")
+        assert response.status_code == status.HTTP_404_NOT_FOUND
+
+    @pytest.mark.parametrize("permission", (Permissions.ADMIN, Permissions.CLUSTERS_READ))
+    @pytest.mark.parametrize("client_id", ("client-1", "client-2", "client-3"))
```
dusktreader, line 142: Why parametrize these tests? I don't think it would add any additional verification to repeat the test 3 times.

fschuch (approved):

jobbergate-api/tests/apps/clusters/test_routers.py
matheushent: the `client_id` is a variable, so, I think it should be parametrized. Consider the following function for example:
```python
def example(client_id: str):
    return client_id.lower()

Comment (matheushent) [reply jobbergate-api/tests/apps/clusters/test_routers.py:142]: since the `client_id` is a variable in this case, I do think it should be parametrized since I can write logic to make the test pass based on the input but the function itself doesn't behave as it should

File: jobbergate-api/jobbergate_api/apps/clusters/routers.py (hunk 1 above)
Comment (matheushent) [reply jobbergate-api/jobbergate_api/apps/clusters/routers.py:87]: I didn't understand actually. Do you mean `ClusterStatusView(instance)` should be enough or `return instance` would work?

The `instance` object has type `ClusterStatus`, which is a SQLAlchemy ORM model. I'm not sure if FastAPI would handle it accordingly. As well as, I don't think it's best practice in terms of coding to return a SQLAlchemy ORM model.

Comment (fschuch) [reply jobbergate-api/jobbergate_api/apps/clusters/routers.py:87]: FastAPI can handle the ORM model internally, other endpoints in the project are coded in this way as well, so I agree with Tucker for consistency

File: jobbergate-api/tests/apps/clusters/test_routers.py (hunk 2 above)
Comment (fschuch) [reply jobbergate-api/tests/apps/clusters/test_routers.py:142]: Both are good points, I see no need to repeat the tests in this case, the three of them will check the same thing. Client id can still be a variable set on the body:

```python
     async def test_get_cluster_status_by_client_id__found(
         self, permission, client, inject_security_header, synth_session
     ):
         client_id = "dummy-client"
```

jobbergate-api/jobbergate_api/apps/clusters/routers.py
matheushent: I understand FastAPI can handle an ORM model internally, however, this pattern is a bit weird to me. It makes useless the concept of validators and serializers then, and if I use somewhere in the code specifically because of a serializer let's say, it would be inconsistent with other endpoints.

That said, the modification was made in [77da40b](https://github.com/omnivector-solutions/jobbergate/pull/591/commits/77da40b5209da556a9dc30fc58d3abd6e38ab6d4).

dusktreader (approved):

PR #607, opened 2024-08-28
[PENG-2342] Jobbergate Agent continually resubmits the same job if the job status update fails

fschuch:
#### What
Create a local cache of job submissions for the jobbergate agent.:
* Check the local cache to see if a job submission was already dispatched to Slurm
* If the job was not already dispatched:
   * submit the job to Slurm
   * Create an entry in the cache to indicate that the job was submitted
* Attempt to update the Job Submission in the Jobbergate API
* If the update was successful delete the entry in the cache

#### Why
We've found a situation where the jobbergate-agent resubmits a job constantly to slurm despite the job having been successfully submitted.
After the Jobbergate Agent successfully submits a pending job submission to slurm, it then attempts to update the status of the job in the Jobbergate API to indicate that the job was submitted. However, if the call to update the job submission in the Jobbergate API fails for any reason, the job_submission will be left in the pending status. This means that in the next cycle of the Jobbergate Agent, it will still see the job as if it had never been submitted and will try to submit the job again.
`Task`: https://app.clickup.com/t/18022949/PENG-2342

---

#### Peer Review

### DIFF
--- a/jobbergate-agent/tests/jobbergate/test_submit.py
+++ b/jobbergate-agent/tests/jobbergate/test_submit.py
@@ -637,6 +637,8 @@ async def test_submit_job_script__raises_exception_if_sbatch_fails(
 @pytest.mark.asyncio
 @pytest.mark.usefixtures("mock_access_token")
 async def test_submit_pending_jobs(
+    tweak_settings,
+    tmp_path,
     dummy_job_script_files,
     dummy_template_source,
     mocker,
@@ -666,6 +668,12 @@ async def test_submit_pending_jobs(
             owner_email="email3@dummy.com",
             job_script={"files": dummy_job_script_files},
         ),
+        PendingJobSubmission(
+            id=4,
+            name="sub4",
+            owner_email="email4@dummy.com",
+            job_script={"files": dummy_job_script_files},
+        ),
     ]
 
     mocker.patch(
@@ -678,7 +686,7 @@ async def test_submit_pending_jobs(
             raise Exception("BOOM!")
         return pending_job_submission.id * 11
 
-    def _mocked_mark_as_submitted(job_submission_id: int, slurm_job_id: int):
+    def _mocked_mark_as_submitted(job_submission_id: int, slurm_job_id: int, slurm_job_data: SlurmJobData):
         if job_submission_id == 2:
             raise Exception("BANG!")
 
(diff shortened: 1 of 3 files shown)


jobbergate-agent/tests/jobbergate/test_submit.py
```diff
@@ -699,7 +707,10 @@ def _mocked_mark_as_submitted(job_submission_id: int, slurm_job_id: int):
 
     test_mapper = manufacture()
 
-    await submit_pending_jobs()
+    with tweak_settings(CACHE_DIR=tmp_path):
+        cached_submission_4 = tmp_path / "4.slurm_job_id"
+        cached_submission_4.write_text("44")
+        await submit_pending_jobs()
```
matheushent, line 713: I guess it's missing a case to test if the cache file is created in case it doesn't exist

jobbergate-agent/tests/jobbergate/test_submit.py
```diff
@@ -714,6 +725,13 @@ def _mocked_mark_as_submitted(job_submission_id: int, slurm_job_id: int):
...
         ]
     )
-    assert mock_mark.call_count == 2
+    assert mock_mark.call_count == 3
+
+    assert cached_submissions[1].exists() is False
+    assert cached_submissions[2].exists() is True
+    assert cached_submissions[2].read_text() == "22"
+    assert cached_submissions[3].exists() is False
+    assert cached_submissions[4].exists() is False
```
dusktreader, line 737: You might consider writing these assertions as:
```suggestion
    assert not cached_submissions[1].exists()
    assert cached_submissions[2].exists()
    assert cached_submissions[2].read_text() == "22"
    assert not cached_submissions[3].exists()
    assert not cached_submissions[4].exists()
```
I think it reads a bit nicer.

dusktreader (approved):
Nice work! I added one very minor comment that you may choose to ignore if you wish.

jobbergate-agent/tests/jobbergate/test_submit.py
fschuch: Thanks for the suggestion. I added some assertions on the cache files to cover this scenario as well. Addressed by [5d46d78](https://github.com/omnivector-solutions/jobbergate/pull/607/commits/5d46d78036c039e69ae44ca33e3bdb8b56e35212).

PR #640, opened 2024-10-25
Tucker/peng 2426  add upload by url endpoints to jobbergate api

dusktreader:
#### What
    PENG-2426 Fixed file uploads with s3

    Jobbergate uses MinIO for file storage internally. The s3 clients use
    environment variables for the AWS secrets. These secrets are bound to
    the MinIO instance. They will not work for s3 in general. When we were
    trying to pull files by URL from s3, our clients were trying to use the
    credentials stored in the environment. These credentials were not known
    by s3.

    Instead of doing all that, we make the assumption that download-by-url
    will require the files to be available publicly. Consequently, we can
    construct a normal https URL to download the file.

    This change applies that logic and fixes the unit tests.

`Task`: https://app.clickup.com/t/18022949/PENG-2426

---

#### Peer Review

### DIFF
--- a/jobbergate-api/Makefile
+++ b/jobbergate-api/Makefile
@@ -3,49 +3,39 @@
 SHELL:=/bin/bash
 PACKAGE_NAME:=jobbergate_api
 
-.PHONY: install
-install:
+install:  ## Install the package locally for development.
 	poetry install
 
-.PHONY: test
-test: install
+test: install  ## Run the unit tests.
 	poetry run pytest
 
-.PHONY: mypy
-mypy: install
+mypy: install  ## Run static type checking on the code base.
 	poetry run mypy ${PACKAGE_NAME} --pretty
 
-.PHONY: lint
-lint: install
+lint: install  ## Run linters over the code base.
 	poetry run ruff check tests ${PACKAGE_NAME}
 
-.PHONY: format
-format: install
+format: install  ## Autoformat the code base.
 	poetry run ruff format tests ${PACKAGE_NAME}
 
-.PHONY: qa
-qa: test mypy lint format
+qa: test mypy lint format  ## Run all quality checks.
 	echo "All quality checks pass!"
 
-.PHONY: local
-local: install
+local: install  ## Run a local dev server.
 	poetry run dev-tools dev-server --port=8000
 
 # To include a message with the generated migration, set the MESSAGE variable in the make command:
 #   $ make db-migration MESSAGE="this migration applies foo to bar"
-.PHONY: db-migrate
-db-migrate: install
+db-migrate: install  ## Create a migration for the database based on the current models.
 	poetry run alembic --config=alembic/alembic.ini revision --autogenerate --message "$(MESSAGE)"
 
 # To override the upgrade target from head, set the UPGRADE_TARGET variable in the make command:
 #   $ make db-upgrade UPGRADE_TARGET=bf1e9e
 UPGRADE_TARGET ?= head
-.PHONY: db-upgrade
-db-upgrade: install
+db-upgrade: install  ## Upgrade the database to a specified target. Defaults to latest.
 	poetry run alembic --config=alembic/alembic.ini upgrade $(UPGRADE_TARGET)
 
-.PHONY: clean
-clean:
+clean:  ## Clean up build artifacts and other junk
 	@find . -iname '*.pyc' -delete
 	@find . -iname '*.pyo' -delete
 	@find . -iname '*~' -delete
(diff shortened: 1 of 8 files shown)


jobbergate-api/jobbergate_api/apps/services.py
```diff
@@ -585,15 +575,14 @@ async def _get_file_data_from_url(self, file_url: AnyUrl) -> io.BytesIO:
                     raise_exc_class=ServiceError,
                     raise_kwargs=dict(status_code=status.HTTP_400_BAD_REQUEST),
                 )
-
-                session = Session()
-                async with session.resource("s3", url_string) as s3:
-                    bucket = await s3.Bucket(bucket_name)
-                    s3_obj = await bucket.Object(key)
-                    await s3_obj.download_fileobj(file_obj)
+                file_url_string = f"https://{bucket_name}.amazonaws.com{key}"
```
matheushent, line 578: we should somehow verify for a trailing slash here in the `key` variable

jobbergate-api/jobbergate_api/apps/services.py
```diff
@@ -585,15 +575,14 @@ async def _get_file_data_from_url(self, file_url: AnyUrl) -> io.BytesIO:
...
+            case "http" | "https":
+                file_url_string = file_url.unicode_string()
             case _:
                 raise ServiceError(f"Unsupported protocol to get file data by url for {file_url}")
 
+        response = httpx.get(file_url_string)
+        file_obj.write(response.content)
```
matheushent, line 585: What do you think in making some assertion here to check if the file can be downloaded?

jobbergate-api/jobbergate_api/logging.py
matheushent, line 1: despite this being a very good addition to the project, I think it should have its own PR to keep this one scoped according to the task

matheushent (changes requested):
I got a few comments for ya. In general, just small stuff

jobbergate-api/Makefile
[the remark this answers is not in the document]
dusktreader: I think it just creates a lot of extra noise. We aren't using Makefiles in a traditional sense, an none of the targets will be used to build files. I would prefer to omit the ugly `.PHONY` markup.

jobbergate-api/jobbergate_api/apps/services.py
dusktreader: We are already enforcing that it's defined. If defined, the path will always have a leading slash due to the way `AnyUrl` parses them.

matheushent: right. Also regarding this line, as mentioned [here](https://app.clickup.com/t/18022949/PENG-2426?comment=90110089671267&threadedComment=90110090042093), it is missing `.s3` before `.amazonaws`

matheushent: > We are already enforcing that it's defined. If defined, the path will always have a leading slash due to the way `AnyUrl` parses them.

but I actually meant a trailing slash, not a leading slash

jobbergate-api/Makefile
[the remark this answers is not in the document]
matheushent: fine, I just think it's a good practice to keep it, being ugly or not

jobbergate-api/jobbergate_api/apps/services.py
dusktreader: Added error handling for the `GET` request in ed5878d.

jobbergate-api/jobbergate_api/logging.py
dusktreader: Removed in commit 7e8ef2e.

I'll add it in a different PR.

jobbergate-api/jobbergate_api/apps/services.py
dusktreader: It's not a valid s3 url if it has a trailing slash, is it? Are you proposing that we strip trailing slashes?

matheushent: I'm actually proposing we don't accept URLs with trailing slash

jobbergate-api/jobbergate_api/apps/services.py
```diff
@@ -585,15 +575,21 @@ async def _get_file_data_from_url(self, file_url: AnyUrl) -> io.BytesIO:
                     raise_exc_class=ServiceError,
                     raise_kwargs=dict(status_code=status.HTTP_400_BAD_REQUEST),
                 )
-
-                session = Session()
-                async with session.resource("s3", url_string) as s3:
-                    bucket = await s3.Bucket(bucket_name)
-                    s3_obj = await bucket.Object(key)
-                    await s3_obj.download_fileobj(file_obj)
+                file_url_string = f"https://{bucket_name}.amazonaws.com{key}"
```
matheushent, line 578: Creating a separated discussion... check my comment [here](https://github.com/omnivector-solutions/jobbergate/pull/640/files#r1817085106) please

fschuch (approved):

jobbergate-api/jobbergate_api/apps/services.py
dusktreader: Jeez, I totally missed it. 🤦🏼

dusktreader: Addressed in commit a8e5887

jobbergate-api/jobbergate_api/apps/services.py
matheushent: just noticed you're calling `raise_for_status`. Disregard the conversation.

PR #683, opened 2025-01-14
feat(api): PENG-2455 create endpoint for fetching timestamps associated to metrics

matheushent:
This PR introduces an endpoint that retrieves the minimum and maximum timestamps of the metrics associated with a specific job submission. The endpoint returns 200 in case of successful request, otherwise 404 in case either the job submission doesn't exist or doesn't have any metrics
associated with it.

### DIFF
--- a/jobbergate-api/tests/apps/job_submissions/test_routers.py
+++ b/jobbergate-api/tests/apps/job_submissions/test_routers.py
@@ -2628,7 +2628,7 @@ async def test_job_submissions_metrics__start_time_less_greater_than_end_time(
         (Permissions.JOB_SUBMISSIONS_READ, 8),
     ],
 )
-async def job_submissions_metrics_timestamps__successful_request(
+async def test_job_submissions_metrics_timestamps__successful_request(
     permission,
     num_rows,
     fill_job_script_data,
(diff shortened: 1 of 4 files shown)


fschuch (approved):

fschuch (changes requested):
Implementation looks good. Please, just add a new entry on the changelog file.

jobbergate-api/tests/apps/job_submissions/test_routers.py
```diff
@@ -2619,3 +2619,157 @@ async def test_job_submissions_metrics__start_time_less_greater_than_end_time(
...
+
+
+@pytest.mark.parametrize(
+    "permission, num_rows",
+    [
+        (Permissions.ADMIN, 3),
+        (Permissions.JOB_SUBMISSIONS_READ, 8),
+    ],
+)
+async def job_submissions_metrics_timestamps__successful_request(
```
dusktreader, line 2631: Why don't these tests start with `test_`? Are they even being run as part of the testing suite?

dusktreader (changes requested):

jobbergate-api/tests/apps/job_submissions/test_routers.py
matheushent: I forgot, thanks.

matheushent:
> Implementation looks good. Please, just add a new entry on the changelog file.

There's already an entry in the changelog under the unreleased section for the whole job metric implementation, so this change is include in it since the whole work is still in a feature branch.

fschuch (approved):

jobbergate-api/tests/apps/job_submissions/test_routers.py
matheushent: check [9f2eaf0](https://github.com/omnivector-solutions/jobbergate/pull/683/commits/9f2eaf02e9dcce057ebc06c15627505180ae8210)

dusktreader (approved):

PR #696, opened 2025-01-31
fix(jobbergate-agent-snap): Fixed problem config hook

dusktreader:
When the `snap set` command is used in quick succession (such as when invoked through the `cloud-init.yaml` of the democluster, the command can fail. This is happening because the snap attempts to restart rapidly when it fails to start. The start will fail because the jobbergate-agent is missing required config.

When the configure hook directs the snap to restart, it may happen while the snap is in a restart cycle. Snap will detect that the restart request happened too soon after a failed restart attempt and the command will fail.

Setting the snap as disabled while the configure hook is running seems to fix this issue.

### DIFF
--- a/jobbergate-agent-snap/hooks/bin/configure
+++ b/jobbergate-agent-snap/hooks/bin/configure
@@ -4,6 +4,7 @@
 import os
 import subprocess
 import sys
+from contextlib import contextmanager
 from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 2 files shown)


fschuch (approved):
Approved. As a suggestion, I'm fine with omitting "jobbergate" when specifying the component in the commit message. Just `fix(agent-snap)`, `fix(api)`, or `fix(cli)` are better to read in my opinion.

dusktreader:
> Approved. As a suggestion, I'm fine with omitting "jobbergate" when specifying the component in the commit message. Just `fix(agent-snap)`, `fix(api)`, or `fix(cli)` are better to read in my opinion.

That's a good observation. Will use that convention going forward.

PR #708, opened 2025-02-11
ci: update test runners to ubuntu 24.04

fschuch:


fschuch:
Ubuntu 20.04 Actions runner image is deprecated: https://github.com/actions/runner-images/issues/11101

PR #713, opened 2025-02-12
fix(agent): solve permission denied when asserting submit dir exist

fschuch:
The previsous approach with `submit_dir.exists()` runs as the agent user, which may face permission errors on the directory (like `/home/fschuch`). It is now verified as the submitter user.

### DIFF
--- a/jobbergate-agent/jobbergate_agent/jobbergate/submit.py
+++ b/jobbergate-agent/jobbergate_agent/jobbergate/submit.py
@@ -5,10 +5,16 @@ import os
 import pwd
 from dataclasses import dataclass
 from pathlib import Path
+from subprocess import CompletedProcess
 from tempfile import TemporaryDirectory
 
 from buzz import DoExceptParams, handle_errors_async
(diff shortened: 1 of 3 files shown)


jobbergate-agent/jobbergate_agent/jobbergate/submit.py
```diff
@@ -184,6 +190,24 @@ def run(self, *args, **kwargs):
         return super().run(*args, **kwargs)
 
 
+def validate_submit_dir(submit_dir: Path, subprocess_handler: SubprocessAsUserHandler) -> None:
```
dusktreader, line 193: We should have unit tests for this function specifically.

dusktreader (changes requested):

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
fschuch: Addressed by [04257e7](https://github.com/omnivector-solutions/jobbergate/pull/713/commits/04257e7c66b728095aa15bbeac2acb0c16b5b88e).

dusktreader (approved):

PR #716, opened 2025-02-12
fix(agent,snap): Fixed the agent auto-update functionality

dusktreader:
Changed the way that agents auto-update if they are using snaps. Instead of using pip to install the target version, the process will use the `refresh` snap command to install the version.

Also returned the auto-update settings to the snap.


fschuch (approved):

fschuch (approved):

dusktreader:
We're going in an entirely new direction. Auto-updating in this form will be removed soon.

PR #722, opened 2025-02-13
fix(agent): Fixed json validation in slurm job data

dusktreader:
Slurm started returning the slurm job state as a list in Slurm 23.11.
In order to support this and older versions of Slurm, the pydantic
schema was updated to handle both.


jamesbeedy (approved):
+1

PR #727, opened 2025-02-14
cherry-pick(jobbergate-api): Removed the sa.Dialect type from alembic migration

jonathanlmd:
Vantage API runs the migrations during the organization creation process. Due to an incompatibility between versions 1.4 and 2.0 of SQLalchmey used by vantage api, it was needed to remove the Dialect type from alembic migration.


fschuch (approved):
Approved. Please, ensure it was covered on main, release/5.5, and release/5.4.

fschuch:
Notice this is duplicated with https://github.com/omnivector-solutions/jobbergate/pull/726, so we can probably close this one.

matheushent:
> Approved. Please, ensure it was covered on main, release/5.5, and release/5.4.

I think this is only relevant to `main` and `release/5.5` (which the original commit comes from).

PR #732, opened 2025-02-14
[PENG-2668] Modify default slurm work dir on Agent

zsinx6:


jobbergate-agent/jobbergate_agent/jobbergate/submit.py
```diff
@@ -244,7 +244,7 @@ async def _reject_handler(params: DoExceptParams) -> None:
         logger.debug(f"Using local slurm user '{username}' for job submission")
         subprocess_handler = SubprocessAsUserHandler(username)
 
-    submit_dir = pending_job_submission.execution_directory or SETTINGS.DEFAULT_SLURM_WORK_DIR
+    submit_dir = pending_job_submission.execution_directory or Path(f"/home/{username}")
```
matheushent, line 247: This assumes that the base home directory is at `/home`, which I think it's not a fair assumption. I'd say we should probably create a setting with this value defaulting to `/home`, so it is customizable if needed.

matheushent (changes requested):

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
fschuch: Following what Matheus suggested, we could keep `SETTINGS.DEFAULT_SLURM_WORK_DIR`, set its default to `"/home/{username}"`, and format it at submission time `SETTINGS.DEFAULT_SLURM_WORK_DIR.format(username=username)`.

jobbergate-cli/jobbergate_cli/subapps/job_submissions/tools.py
```diff
@@ -74,7 +74,7 @@ def __post_init__(self):
         self.cluster_name = _map_cluster_name(self.jg_ctx, self.cluster_name)
 
         if self.execution_directory is None:
-            self.execution_directory = Path.cwd()
+            self.execution_directory = Path.home()
```
fschuch, line 77: The change on the agent side looks good, but there is no need to change it on the cli, `cwd` is the expected here.

fschuch (changes requested):

fschuch:
Please, add a new entry on the changelog.

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
dusktreader: No, we should remove the `DEFAULT_SLURM_WORK_DIR`, I think. We can use `Path.home()` instead. This should always return a valid directory according to the Python docs and POSIX specs:

https://docs.python.org/3/library/pathlib.html#pathlib.Path.home
https://docs.python.org/3/library/os.path.html#os.path.expanduser
https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html (see entry for Other Environment Variables > HOME)

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
```diff
@@ -244,7 +244,9 @@ async def _reject_handler(params: DoExceptParams) -> None:
         logger.debug(f"Using local slurm user '{username}' for job submission")
         subprocess_handler = SubprocessAsUserHandler(username)
 
-    submit_dir = pending_job_submission.execution_directory or SETTINGS.DEFAULT_SLURM_WORK_DIR
+    submit_dir = pending_job_submission.execution_directory or Path(
+        SETTINGS.DEFAULT_SLURM_WORK_DIR.format(username=username)
```
dusktreader, line 248: As I said in another comment, we should remove the `DEFAULT_SLURM_WORK_DIR` entirely and use `Path.home()` here. We don't actually have a use-case that I know of for using `DEFAULT_SLURM_WORK_DIR`. Simpler is better when we can have it.

jobbergate-agent/jobbergate_agent/settings.py
```diff
@@ -29,7 +29,7 @@ class Settings(BaseSettings):
     SBATCH_PATH: Path = Path("/usr/bin/sbatch")
     SCONTROL_PATH: Path = Path("/usr/bin/scontrol")
     X_SLURM_USER_NAME: str = "ubuntu"
-    DEFAULT_SLURM_WORK_DIR: Path = Path("/tmp")
+    DEFAULT_SLURM_WORK_DIR: str = "/home/{username}"
```
dusktreader, line 32: See my comment above.

dusktreader (changes requested):

CHANGELOG.md
```diff
@@ -11,6 +11,46 @@ Refer to [changes](./changes) directory for unreleased changes.
 
 <!-- towncrier release notes start -->
 
+# [5.5.0](https://github.com/omnivector-solutions/jobbergate/releases/tag/5.5.0) - 2025-02-18
```
fschuch, line 14: Please, revert the changes on this file. We just need it at release time, not on a feature branch.

fschuch: Notice we also need the news fragments back

fschuch (changes requested):

CHANGELOG.md
dusktreader: @fschuch you may need to make the requisite changes and push this PR forward.

fschuch: Addressed by [4adadff](https://github.com/omnivector-solutions/jobbergate/pull/732/commits/4adadff00887bbf32d8a8e59e98b64ace8f399e8).

jobbergate-cli/jobbergate_cli/subapps/job_submissions/tools.py
fschuch: Addressed by [5b2c67c](https://github.com/omnivector-solutions/jobbergate/pull/732/commits/5b2c67cc36f1921a61b226570ea1a3275a0d414c).

jobbergate-agent/jobbergate_agent/jobbergate/submit.py
fschuch: We got to an agreement on the comment https://github.com/omnivector-solutions/jobbergate/pull/732#discussion_r1958203915.

fschuch (approved):

PR #733, opened 2025-02-14
chore(snap,agent,ci): Minor quality-of-life updates

dusktreader:
- Added help text for some Makefile items that were missing them
- Improved clean target for snap Makefile
- Updated the CI to use default snap name

### DIFF
--- a/jobbergate-agent-snap/Makefile
+++ b/jobbergate-agent-snap/Makefile
@@ -16,16 +16,16 @@ format: ## Run the formatter
 	tox -e format
 
 .PHONY: build
-build: ## Build the snap
-	@git rev-parse --short HEAD > sha/commit-sha.txt
-	@snapcraft $(ARGS)
+build: clean ## Build the snap
+	@snapcraft -v
+	mv *.snap jobbergate-agent.snap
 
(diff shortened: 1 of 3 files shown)


jobbergate-agent-snap/Makefile
```diff
@@ -16,16 +16,16 @@ format: ## Run the formatter
 	tox -e format
 
 .PHONY: build
-build: ## Build the snap
-	@git rev-parse --short HEAD > sha/commit-sha.txt
-	@snapcraft $(ARGS)
+build: clean ## Build the snap
+	@snapcraft -v
+	mv *.snap jobbergate-agent.snap
```
zsinx6, line 21: Can you explain this part?

dusktreader: Sure!

I made it dependent on `clean` so that the build environment would be fresh and any leftover `.snap` artifacts would be removed.

The command `snapcraft -v` builds the snap using verbose mode. (I can't imagine not wanting to see the output because if something goes wrong, you have to run it again in verbose mode anyway. Additionally, the CI was already using verbose mode. Just seemed like the right move.

PR #739, opened 2025-03-07
fix(agent): update the `validate_job_state` validator to support data_parser 0.0.40

matheushent:
This PR modifies the validator function `validate_job_state` from the `SlurmJobData` model in order to support the data_parser 0.0.40.

Essentialy, in Slurm 24.11, the `--json` flag from the `scontrol show job` command uses the data_parser 0.0.40. According to the official [documentation](https://slurm.schedmd.com/job_state_codes.html#overview), the job flags can be returned alongside the job state, which can make the `job_state` key to have multiple keys.

### DIFF
--- a/changes/agent/739.fixed.md
+++ b/changes/agent/739.fixed.md
@@ -1,0 +1,6 @@
+Updated the `validate_job_state` validator to support data_parser 0.0.40.
+
+In Slurm 24.11, the `--json` flag from the `scontrol show job` command uses the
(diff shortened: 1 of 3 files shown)


fschuch (changes requested):
The code looks good, but please, add a news fragment so we can track it on new releases.

matheushent:
> The code looks good, but please, add a news fragment so we can track it on new releases.

check [235c9ff](https://github.com/omnivector-solutions/jobbergate/pull/739/commits/235c9ff5603c2ac3a65faa0a9a9ca4b4e73f616a) please.

PR #749, opened 2025-03-18
fix(agent): [PENG-2912] constrain the measurements returned by the `fetch_influx_measurements` function.

matheushent:
This PR modifies the behaviour of the `fetch_influx_measurements` function in order to only return the measurements defined in the `INFLUXDB_MEASUREMENT` constant. This is necessary becase some Slurm clusters may have other measurements, which leads to an error when the agent sends the metrics to the API.


fschuch (approved):
Approved. Please, consider adding the PR label `backport/main` if you need to git cherry-pick it.

PR #756, opened 2025-04-08
fix(agent): ASP-5933 implement fetch_paginated_result for active and pending job submissions

fschuch:
Pending and active submissions can now handle multiple pages per cycle. A new function `fetch_paginated_result` was added to cover this.


jobbergate-agent/tests/jobbergate/test_update.py
```diff
@@ -533,7 +537,7 @@ async def test_fetch_influx_data__success_with_all_None(mocked_influxdb_client:
 @pytest.mark.parametrize(
     "time, host, step, task",
     [
-        tuple(random.randint(1, 100) if i not in combination else None for i in range(4))
+        tuple(1 if i not in combination else None for i in range(4))
```
fschuch, line 540: A random parameter made it challenging to work on the project using vscode. I pinned it to 1 for now, since the value doesn't seem relevant on this test. We may revisit it in the future.

PR #775, opened 2025-06-03
chore(api): updated dependencies

fschuch:


fschuch:
> please, include in the first commit message a more descriptive body. Specifically, I think it'd be good to detail which dependencies have been updated from version x to y.
> 
> Also, it seems to me the two commits here should be in different PRs.

Thanks for the comments. The 

> please, include in the first commit message a more descriptive body. Specifically, I think it'd be good to detail which dependencies have been updated from version x to y.
> 
> Also, it seems to me the two commits here should be in different PRs.

Thanks for your comments. I added more details on the commit message with the main version updates and the links to the dependabot alerts they solve. I double-checked the alerts are only visible to maintainers on the project.

I also addressed the CI fix in dedicated PR: https://github.com/omnivector-solutions/jobbergate/pull/777
I reverted the commit here and rebased the branch, so we are all good.

matheushent (approved):

PR #829, opened 2025-07-21
feat(agent): implement job cancellation functionality

fschuch:
Notice this PR covers Agent work and it is going into a feature branch where integration tests and another round of code review are planned before going in fact to the main branch.


julianaklulo (approved):
Nice work!

+1

PR #833, opened 2025-08-08
feat(jobbergate-api): PENG 3064 Keycloak migration to 26

jonathanlmatos:


fschuch (changes requested):
Changes approved. Please, just create a changelog fragment with `make changes` (use PR number when it asks for issue number)

